PR #56, opened 2016-08-08
Fixes #28 Add support for auth method enumeration

### LINKED ISSUE #28: Add support for auth method enumeration
Compliments of @iamlei

Would be nice to have the enumeration component support auth methods allowed (key, pass, etc..)

jinankjain:
I have added auth enumerations using net-ssh. Have a look at the patch and any comments? @claudijd


claudijd:
Oh cool, I didn't realize @auth_methods was being populated properly.  In the other example we discussed, I figured it would require a monkey patch.  Thanks, going to land this now.

Issue #27, opened 2016-02-23
Add support for scanning multiple IPs


claudijd:
Compliments of @leighalytle.

It should support scanning more than one host at a time.

1.) -t switch should support NMAP syntax targeting
2.) should add a -f switch to accept targeting info via file (should support NMAP style syntax)

claudijd:
Requirements:

1.) Need a native mechanism to detect open ports
2.) Need better exception/error handling/output
3.) Need parsing logic and classes for NMAP-style targeting info

claudijd:
+1 (ran into a need for this again today)

claudijd:
4.) Have a way to detect SSH protocol (go a step beyond just knowing a service has port 22 open before we attempt to scan)

jinankjain:
@claudijd When doing a request over multiple IPs do you want result for each ip in different array or final result to get combined in one result array

claudijd:
@jinankjain I would like the resultant output of a multi-ip scan to be a single JSON document and an array of what we normally get from a scan seems like the right fit.

jinankjain:
@claudijd so results for all the IP would be combined that single json array?

claudijd:
@jinankjain is making some progress on IP comma separated targeting in #65.

Other syntax we should consider supporting could include....

192.168.1.0/24 (== 192.168.1.1 thru 192.168.1.255)
192.168.1.1-255 (== 192.168.1.1 thru 192.168.1.255)
192.168.1.1,2 (== 192.168.1.1 && 192.168.1.2)

To gain the above, we'll probably need to implement some sort of IP parsing logic (at least for the CIDR notation).  Consider Ruby IPAddr and CIDR gems to help with that.  Might consider some customer parsing logic for the others.

claudijd:
Perhaps we should also have a command-line switch to load these from a file.  Looks like -f it still open.  Might also consider having a -o flag for writing our JSON documents to disk.

claudijd:
@jinankjain yes, a single JSON array.

jinankjain:
@claudijd I have implemented -f for loading targets via file but waiting for #65 to land so that could use that piece of code :smile:

claudijd:
Re-opening this issue, because it includes more than what was added in https://github.com/mozilla/ssh_scan/pull/65 I also realized we might consider updating the -h output to include an example of multi-ip usage.

claudijd:
[...]

PR #65, opened 2016-08-10
Multi IP Support comma seperated

jinankjain:
@claudijd  Syntax looks like this ./bin/ssh_scan -t github.com,128.237.153.145

### DIFF
--- a/bin/ssh_scan
+++ b/bin/ssh_scan
@@ -8,7 +8,7 @@ require 'optparse'
 
 #Default options
 options = {
-  :target => nil,
+  :targets => [],
   :port => 22,
   :policy => File.expand_path("../../policies/mozilla_modern.yml", __FILE__),
   :unit_test => false
@@ -17,9 +17,9 @@ opt_parser = OptionParser.new do |opts|
   opts.banner = "ssh_scan v#{SSHScan::VERSION} (https://github.com/mozilla/ssh_scan)\n\n" +
                 "Usage: ssh_scan [options]"
 
-  opts.on("-t", "--target [IP/Hostname]",
+  opts.on("-t", "--target [IP/Hostname]", Array,
           "IP/Hostname (IPv4/IPv6/FQDNs)") do |ip|
-    options[:target] = ip
+    options[:targets] = ip
   end
 
   opts.on("-p", "--port [PORT]",
@@ -59,16 +59,18 @@ end
 
 opt_parser.parse!
 
-if options[:target].nil?
+if options[:targets].nil?
   puts opt_parser.help
   puts "\nReason: no target specified"
   exit
 end
 
-unless options[:target].ip_addr? || options[:target].fqdn?
-  puts opt_parser.help
-  puts "\nReason: #{options[:target]} is not a valid target"
-  exit
+options[:targets].each do |target|
+  unless target.ip_addr? || target.fqdn?
+    puts opt_parser.help
+    puts "\nReason: #{options[:targets]} is not a valid target"
+    exit
+  end
 end
 
 unless (0..65535).include?(options[:port])
@@ -87,12 +89,14 @@ options[:policy_file] = SSHScan::Policy.from_file(options[:policy])
 
 # Perform scan and get results
 scan_engine = SSHScan::ScanEngine.new()
-result = scan_engine.scan(options)
+results = scan_engine.scan(options)
 
-puts JSON.pretty_generate(result)
+puts JSON.pretty_generate(results)
 
-if result["compliance"] && result["compliance"][:compliant] == false
-  exit 1 #non-zero means a false
-else
-  exit 0 #non-zero means pass
+results.each do |result|
+  if result["compliance"] && result["compliance"][:compliant] == false
+    exit 1 #non-zero means a false
+  else
+    exit 0 #non-zero means pass
+  end
 end
--- a/lib/ssh_scan/scan_engine.rb
+++ b/lib/ssh_scan/scan_engine.rb
@@ -6,49 +6,56 @@ module SSHScan
   class ScanEngine
 
     def scan(opts)
-      target = opts[:target]
+      targets = opts[:targets]
       port = opts[:port]
       policy = opts[:policy_file]
 
       # Connect and get results (native)
-      client = SSHScan::Client.new(target, port)
-      client.connect()
-      result = client.get_kex_result()
+      result = []
+      targets.each_with_index do |target, index|
+        client = SSHScan::Client.new(target, port)
+        client.connect()
+        result.push(client.get_kex_result())
 
-      # Connect and get results (Net-SSH)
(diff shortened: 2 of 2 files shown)


bin/ssh_scan
```diff
@@ -65,10 +65,12 @@ if options[:target].nil?
...
-unless options[:target].ip_addr? || options[:target].fqdn?
-  puts opt_parser.help
-  puts "\nReason: #{options[:target]} is not a valid target"
-  exit
+for ip in options[:target]
+  unless ip.ip_addr? || ip.fqdn?
+    puts opt_parser.help
+    puts "\nReason: #{options[:target]} is not a valid target"
+    exit
+  end
```
claudijd: for loops in Ruby are not preferred based on the style guide (https://github.com/bbatsov/ruby-style-guide)

A more preferred way to express this would be like so...

    options[:target].each do |target|
      unless target.ip_addr? || target.fqdn?
        puts opt_parser.help
        puts "\nReason: #{options[:target]} is not a valid target" 
        exit
      end
    end

bin/ssh_scan
```diff
@@ -8,7 +8,7 @@ require 'optparse'
 
 #Default options
 options = {
-  :target => nil,
+  :target => [],
```
claudijd: maybe the name should change to "targets" since it's no longer singular?

lib/ssh_scan/scan_engine.rb
```diff
@@ -11,44 +11,49 @@ def scan(opts)
...
-      result['auth_methods'] = auth_session.allowed_auth_methods
-      host_key = net_ssh_session.host_keys.first
-      net_ssh_session.close
-
-      # only supporting RSA for the moment
-      unless OpenSSL::PKey::RSA
-        raise "Unknown host key type, need to add this host_key type"
+      i = 0
+      result = []
+      while i<target.length do
```
claudijd: while loops are also non-preferred in ruby, generally interators are preferred.

Recommended change would be...

target.each_with_index do |target, i|
....
end

    2.2.3 :001 > target = ["192.168.1.1]
    2.2.3 :002"> ^C
    2.2.3 :002 > target = ["192.168.1.1"]
     => ["192.168.1.1"] 
    2.2.3 :003 > target = ["192.168.1.1", "192.168.1.2"]
     => ["192.168.1.1", "192.168.1.2"] 
    2.2.3 :004 > target.each_with_index do |target, i|
    2.2.3 :005 >     puts target
    2.2.3 :006?>     puts i
    2.2.3 :007?>   end
    192.168.1.1
    0
    192.168.1.2
    1

lib/ssh_scan/scan_engine.rb
```diff
@@ -11,44 +11,49 @@ def scan(opts)
...
-      net_ssh_session.close
-
-      # only supporting RSA for the moment
-      unless OpenSSL::PKey::RSA
-        raise "Unknown host key type, need to add this host_key type"
+      i = 0
+      result = []
+      while i<target.length do 
+        puts target[i]
+        client = SSHScan::Client.new(target[i], port)
```
claudijd: target[i] in this context will just change to target if you use the iterator strategy above.

lib/ssh_scan/scan_engine.rb
```diff
@@ -11,44 +11,49 @@ def scan(opts)
...
+        auth_session = Net::SSH::Authentication::Session.new(net_ssh_session, :auth_methods => ["none"])
+        auth_session.authenticate("none", "test", "test")
+        result[i]['auth_methods'] = auth_session.allowed_auth_methods
+        host_key = net_ssh_session.host_keys.first
+        net_ssh_session.close
+
+        # only supporting RSA for the moment
+        unless OpenSSL::PKey::RSA
+          raise "Unknown host key type, need to add this host_key type"
+        end
```
claudijd: This code recently changed, might want to rebase against master to pull them in.

lib/ssh_scan/scan_engine.rb
```diff
@@ -11,44 +11,49 @@ def scan(opts)
...
+          "sha1" => fingerprint_sha1,
+          "sha256" => fingerprint_sha256,
+        }
+
+        # If policy defined, then add compliance results
+        unless policy.nil?
+          policy_mgr = SSHScan::PolicyManager.new(result[i], policy)
+          result[i]['compliance'] = policy_mgr.compliance_results
+        end
+        i += 1
```
claudijd: this line will go away if you use the iterator strategy

lib/ssh_scan/scan_engine.rb
claudijd: It might also be worth-while to factor out the working part of this code into a method and then just call the method in the iterator.

Example:

    def do_a_scan(target)
       ...code...
    end

    targets.each do |target|
      do_a_scan(target)
    end

It will make it easier to follow without having to embed all the working logic and the iterators together.

bin/ssh_scan
```diff
@@ -91,8 +93,10 @@ result = scan_engine.scan(options)
...
-if result["compliance"] && result["compliance"][:compliant] == false
-  exit 1 #non-zero means a false
-else
-  exit 0 #non-zero means pass
+for res in result
+  if res["compliance"] && res["compliance"][:compliant] == false
+    exit 1 #non-zero means a false
+  else
+    exit 0 #non-zero means pass
+  end
```
claudijd: Same here with the for loops...

results.each do |result|
   #stuff
end

claudijd:
This PR is targeted to make some progress on #27

claudijd:
@jinankjain thanks for the PR, I made some suggestions on how it can be improved before it's landed.

jinankjain:
@claudijd Thanks For suggesting me those changes this is my first time working on a project involving Ruby, I was not aware about such rules, will keep in mind from next time :smile:  
I will update this PR soon

claudijd:
@jinankjain no problem at all, really glad you're working on this, you're doing a great job!

jinankjain:
@claudijd Any comments about this new patch?

claudijd:
@jinankjain It looks like there is a bug in the code based on the CI build here (https://travis-ci.org/mozilla/ssh_scan/jobs/151415302)  The output is this...

"SyntaxError: /home/travis/build/mozilla/ssh_scan/lib/ssh_scan/scan_engine.rb:62: syntax error, unexpected end-of-input, expecting keyword_end"

I believe this is due to a missing "end" on line 58 (https://github.com/jinankjain/ssh_scan/blob/17f9a84b961e157dd90384afe5be73c72f0be868/lib/ssh_scan/scan_engine.rb#L58)

claudijd:
once you push the fix for that, the build should return to passing and I'd be happy to land it.

jinankjain:
@claudijd Fixed it 😄

claudijd:
@jinankjain landed, thank you!

PR #70, opened 2016-08-15
removed extra space which was resulting is server panic

anantshri:
Extra space in list of entries to be added was present if someone copy pastes the list in sshd_config file in hope of getting the right configuration this results in a ssh server launch failure.

Similar space in list of stuff to removed doesn't matters as they will anyways be required to be removed manually.

Hope this helps


claudijd:
@anantshri thanks for that, anything that makes it easier for people to implement the recommendations is great!

PR #76, opened 2016-08-15
Nmap style IP range

jinankjain:


jinankjain:
@claudijd Other two syntax requires custom parser to parse those IP range and comma separated is not possible because are using comma to support multiple IP's So for the time being we can supports these syntax

claudijd:
@jinankjain sounds good, I made a slight tweak when landing to the gemfile, but otherwise good.  This also pushes more on the issue of being able to have fast port/protocol/timeout capabilities.  Thanks!

PR #79, opened 2016-08-15
Redirect Ouput JSON to some file

jinankjain:


claudijd:
@jinankjain landed, thanks!

PR #85, opened 2016-08-15
Build docker container in travisci

jvehent:
If this works, we could add automated publication of the container in the mozilla org on dockerhub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM ruby
+MAINTAINER Jonathan Claudius
+ENV PROJECT=github.com/mozilla/ssh_scan
+
+RUN mkdir /app
+ADD . /app
+WORKDIR /app
+
+RUN gem install bundler
+RUN bundle install
+CMD /app/bin/ssh_scan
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ruby
+MAINTAINER Jonathan Claudius
+ENV PROJECT=github.com/mozilla/ssh_scan
+
+RUN mkdir /app
+ADD . /app
+WORKDIR /app
+
```
claudijd: I'm wondering if adding this on line 9 would do the trick...

rvm --create --ruby-version use ruby-$(ruby -e 'puts RUBY_VERSION')@ssh_scan

jvehent: Unfortunately, the `ruby` container doesn't have `rvm` installed.
`/bin/sh: 1: rvm: not found`

claudijd:
Looks like it's not just travis (https://github.com/mozilla/ssh_scan/pull/87)

claudijd:
@jvehent the recently merged changes into this PR will give us the desired build behavior.

The only thing that is missing, I think, is the publishing step to docker hub.

Dockerfile
claudijd: decided to nix rvm from the docker deployment

Issue #64, opened 2016-08-10
Add a timeout capability and error handling/output when we can't scan an endpoint


claudijd:
This addition would have both a default and a configurable command-line switch to define how long we should hang open waiting for a connection to start before we give up and timeout the scan.

This addition would also need to define some JSON output that we would send when we resulted in an error state.  We should also specify some error text and a reason for why the scan didn't complete.

    {
      "ssh_scan_version": "0.0.7",
      "hostname": "",
      "ip": "192.168.1.1",
      "port": 22,
      "error": "SSHScan::Timeout: unable to reach service due to timeout"
    }

jinankjain:
@claudijd This could be closed now.

claudijd:
@jinankjain thanks for the reminder, always glad to take things off the TODO list.

PR #92, opened 2016-08-16
add timeout options

agaurav77:
With respect to #64. 

Please indicate things that should be changed in the PR, if any. Thanks!

### DIFF
--- a/lib/ssh_scan/client.rb
+++ b/lib/ssh_scan/client.rb
@@ -2,11 +2,13 @@ require 'socket'
 require 'ssh_scan/constants'
 require 'ssh_scan/protocol'
 require 'ssh_scan/banner'
+require 'ssh_scan/error'
 
 module SSHScan
   class Client
-    def initialize(target, port)
+    def initialize(target, port, timeout = 3)
       @target = target
+      @timeout = timeout
 
       if @target.ip_addr?
         @ip = @target
@@ -21,13 +23,31 @@ module SSHScan
     end
 
     def connect()
-      @sock = TCPSocket.new(@ip, @port)
-      @raw_server_banner = @sock.gets.chomp
-      @server_banner = SSHScan::Banner.read(@raw_server_banner)
-      @sock.puts(@client_banner.to_s)
+      begin
+        @sock = Socket.tcp(@ip, @port, connect_timeout: @timeout)
+      rescue Errno::ETIMEDOUT => e
+        @error = SSHScan::Error::ConnectTimeout.new(e.message)
+        @sock = nil
+      else
+        @raw_server_banner = @sock.gets.chomp
+        @server_banner = SSHScan::Banner.read(@raw_server_banner)
+        @sock.puts(@client_banner.to_s)
+      end
     end
 
     def get_kex_result(kex_init_raw = @kex_init_raw)
+      # Common options for all cases
+      result = {}
+      result[:ssh_scan_version] = SSHScan::VERSION
+      result[:hostname] = @target.fqdn? ? @target : ""
+      result[:ip] = @ip
+      result[:port] = @port
+
+      if !@sock
+        result[:error] = @error
+        return result
+      end
+
       @sock.write(kex_init_raw)
       resp = @sock.read(4)
       resp += @sock.read(resp.unpack("N").first)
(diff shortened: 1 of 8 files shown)


lib/ssh_scan/error.rb
```diff
@@ -0,0 +1,10 @@
+module SSHScan
+  class Error < Exception
```
claudijd: Maybe we should be more explicit here about what type of error.

What about calling it "ConnectTimeout"?

lib/ssh_scan/scan_engine.rb
```diff
@@ -9,13 +10,16 @@ def scan(opts)
...
-        client = SSHScan::Client.new(target, port)
+        client = SSHScan::Client.new(target, port, timeout)
         client.connect()
-        result.push(client.get_kex_result())
+        kex_result = client.get_kex_result()
+        result.push(kex_result)
+        next if kex_result.include?(:error)
 
         # Connect and get results (Net-SSH)
         net_ssh_session = Net::SSH::Transport::Session.new(target, :port => port)
```
claudijd: Can we also add a connect timeout for this too?

We might also need to catch the net-ssh specific exception for connect timeouts and re-raise the custom exception you created

lib/ssh_scan/client.rb
```diff
@@ -21,13 +22,31 @@ def initialize(target, port)
     end
 
     def connect()
-      @sock = TCPSocket.new(@ip, @port)
-      @raw_server_banner = @sock.gets.chomp
-      @server_banner = SSHScan::Banner.read(@raw_server_banner)
-      @sock.puts(@client_banner.to_s)
+      begin
+        @sock = Socket.tcp(@ip, @port, connect_timeout: @timeout)
+      rescue Exception => e
```
claudijd: let's be more explicit about the exception type so we're not swallowing all exceptions from this call

Easiest way to test is to point it at something without a begin block around it and seeing what exception it will throw.

lib/ssh_scan/scan_engine.rb
agaurav77: You mean for the Net-SSH connect?

claudijd: yes

lib/ssh_scan/error.rb
claudijd: We might also consider adding a name space here like...

    module SSHScan
      module Exceptions
        class ConnectTimeout
          ....
        end
      end
    end

claudijd:
@agaurav77 some of the dependant code on this PR changed in master to fix a bug, try rebasing to see if that fixes it's merge-ability.

jinankjain:
@agaurav77 Are you working on rebasing the PR?

agaurav77:
@jinankjain yes, I just got stuck with some college work. I'll rebase over the weekend.

jinankjain:
@agaurav77 I tried to rebase but there are still some errors but I think those require some changes in test cases @claudijd What do you say ?

agaurav77:
@claudijd @jinankjain This should be good to go, thanks for your patience!

Actually, I was using a Socket.tcp instead of TCPSocket.new, and some tests were TCPSocket.new specific, so they were failing. Updated the tests.

lib/ssh_scan/error/closed_connection.rb
```diff
@@ -0,0 +1,9 @@
+module SSHScan
+  module Error
+    class ClosedConnection < Exception
+			def to_s
```
jinankjain: There is some indentation error @agaurav77

lib/ssh_scan/error/connect_timeout.rb
```diff
@@ -0,0 +1,12 @@
+module SSHScan
+  module Error
+    class ConnectTimeout < Exception
+      def initialize(message)
+        @message = message
+      end
+			def to_s
```
jinankjain: It has the same problem as above

lib/ssh_scan/error/closed_connection.rb
agaurav77: Fixed, thanks!

lib/ssh_scan/error/connect_timeout.rb
agaurav77: Done

claudijd:
Looks good!

    $ ./bin/ssh_scan -t 192.168.10.20
    [
      {
        "ssh_scan_version": "0.0.9",
        "hostname": "",
        "ip": "192.168.10.20",
        "port": 22,
        "error": "ConnectTimeout: Operation timed out - user specified timeout"
      }
    ]

claudijd:
Thanks @agaurav77 and @jinankjain !

PR #96, opened 2016-08-17
Custom Parser

jinankjain:
I had tried building some custom parser

### DIFF
--- a/lib/string_ext.rb
+++ b/lib/string_ext.rb
@@ -1,4 +1,5 @@
 require 'ipaddr'
+require 'netaddr'
 
 # Extend string to include some helpful stuff
 class String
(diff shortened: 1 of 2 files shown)


lib/string_ext.rb
```diff
@@ -22,6 +23,16 @@ def resolve_fqdn
...
 
+  def resolve_ip
+    ip_ranges = self.split('-')
+    lower = NetAddr::CIDR.create(ip_ranges[0])
+    temp = ip_ranges[0].split('.')
+    temp[3] = ip_ranges[1]
+    upper = NetAddr::CIDR.create(temp.join("."))
+    ip_ranges = NetAddr.range(lower, upper,:Inclusive => true)
+    return ip_ranges
+  end
```
claudijd: I think this parser should be it's own class and have associated unit-tests to prove that it covers all the scenarios we expect.

I would expect files like these to be created for this feature....

https://github.com/mozilla/ssh_scan/tree/master/lib/ssh_scan/target_parser.rb
https://github.com/mozilla/ssh_scan/tree/master/spec/ssh_scan/target_parser_rspec.rb

Once you have the specs written, you can validate your implementation by running "rspec spec/ssh_scan/target_parser_rspec.rb" from the dev root.  See other example of this for guidance, and ask me any questions you have around it.

jinankjain:
@claudijd Thanks for the advice I would this patch soon 😄

PR #97, opened 2016-08-17
Multi Threaded Architecture

jinankjain:


claudijd:
This code is getting a little unwieldy and I've run into threading issues when running this command...

 $ ./bin/ssh_scan -t github.com,ssh.mozilla.com,192.168.10.10
/Users/jclaudius/code/ssh_scan/lib/ssh_scan/scan_engine.rb:31:in `block (2 levels) in scan': undefined method `[]=' for nil:NilClass (NoMethodError)

I think we might consider creating a method that does a scan of a single host and just threading that to avoid some of the ever growing logic here.

Like this....

    def scan_ip(...)
       #do individual scan
    end

    def scan(...)
        @results = []
       threads << Thread.new {
          @results <<  scan_ip(...)
       }
    end

claudijd:
I can help with this refactoring if you're interested.

jinankjain:
Yeah sure :)

claudijd:
ok, I'll work on getting that refactored and will send a PR shortly

claudijd:
Refactored it a fair amount, didn't add any threading since you planning to do that, but this change should make it easier to avoid threading bugs. https://github.com/mozilla/ssh_scan/pull/98

jinankjain:
@claudijd Updated the PR with threading logic

claudijd:
@jinankjain thanks, works great now, no more threading errors when running the command I had issues with before :)

jinankjain:
@claudijd It's kind of surprising to me, previous implementation was also working fine for me but why not for you?

claudijd:
@jinankjain previous implementation was inserting contents into an array that wasn't built yet, so depending on speed of scan (I had a remote and a local in mine) you'd end up with the equivalent of an out of bound error when referencing the index on the array.  The new code appends results as they complete and order is irrelevant.  I think that's the key factor, but I do like the way the new logic is easier to read.

PR #110, opened 2016-08-19
Added Timeout

jinankjain:
I tried to rebase that patch


jinankjain:
@claudijd Have a look at this new patch rewritten for timeout?

Issue #100, opened 2016-08-17
Extend port switch to handle multiple ports


claudijd:
If we know we host ssh endpoints on mixed ports (say 22 and 2222) we might want to scan both port 22 and 2222 in a single scan.

Expected usage would be like...

scan_ssh -f hosts.txt -p 22,2222

If #99 existed, a socket specification in the file would override the port-based "bruteforce".

jinankjain:
#160  Fixes #100 too.

Issue #99, opened 2016-08-17
Extend input file capability to support custom ports


claudijd:
Basically, we might have a list of known ssh endpoints, but they might be running on different ports.

if so, the file input format could support port specification like so...

**BEFORE**

192.168.1.1
192.168.1.2

**AFTER**

192.168.1.1
192.168.1.2:2222

No port specification in command-line or file implies a default of port 22.

Issue #101, opened 2016-08-17
Support output as input for re-scans


claudijd:
Say you did a scan of 192.168.1.1 and 192.168.1.2 and someone shows you a JSON output file from ssh_scan.  It would be cool if you could just perform a rescan of those targets using the output file and get an update without having to build your own host-list.

Example usage would be like this....

**Initial Scan**
ssh_scan -t 192.168.1.1,192.168.1.2 -o output.json

**Rescan using old output**
ssh_scan --from_json output.json -o rescan_output.json

PR #111, opened 2016-08-21
Add support for custom ports in input file and support rescans

royalharsh:
```bash
```
Should fix https://github.com/mozilla/ssh_scan/issues/100 because now you can specify custom ports in the file. Am I correct @claudijd ?

Now we can use 
```bash
```

### DIFF
--- a/bin/ssh_scan
+++ b/bin/ssh_scan
@@ -6,11 +6,13 @@ $:.unshift File.join(File.dirname(__FILE__), "../lib")
 require 'ssh_scan'
 require 'optparse'
 require 'netaddr'
+require 'rubygems'
+require 'json'
 
 #Default options
 options = {
   :targets => [],
-  :port => 22,
+  :ports => [],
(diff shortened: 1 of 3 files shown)


bin/ssh_scan
```diff
@@ -35,8 +35,15 @@ opt_parser = OptionParser.new do |opts|
       exit
     end
     File.open(file).each do |line|
-      line.chomp.split(',').each do |ip|
+      line.chomp.split(',').each do |ips|
```
jinankjain: Naming variables are getting really confusing. So after splitting about comma we would get a socket, so instead  of ips you could name it socket.

bin/ssh_scan
```diff
@@ -35,8 +35,15 @@ opt_parser = OptionParser.new do |opts|
...
+        ip_arr = ips.chomp.split(':')
```
jinankjain: Instead of getting things into array you could do

ip, port = socket.chomp.split(':')

and then check if port is nil then put the default 22 port

PR #112, opened 2016-08-21
Fixes #104

### LINKED ISSUE #104: Define default client key_init message as an object rather than raw blob
This is really just to make sure we're being clear about what's being sent in the initial client key init message and leaving it open to be expanded upon in the future.  The string blob isn't all that transparent.

Make the following definition manually defined as a SSHScan::KeyExchangeInit object...

https://github.com/mozilla/ssh_scan/blob/master/lib/ssh_scan/constants.rb#L8-L21

Example:

    default_key_init = SSHScan::KeyExchangeInit.new()
    ...set all attributes manually...

jinankjain:
@claudijd How about this one?

### DIFF
--- a/lib/ssh_scan/constants.rb
+++ b/lib/ssh_scan/constants.rb
@@ -1,11 +1,14 @@
 require 'string_ext'
 require 'ssh_scan/banner'
+require 'ssh_scan/protocol'
 
 module SSHScan
   module Constants
     DEFAULT_CLIENT_BANNER = SSHScan::Banner.new("SSH-2.0-ssh_scan")
     DEFAULT_SERVER_BANNER = SSHScan::Banner.new("SSH-2.0-server")
-    DEFAULT_KEY_INIT_RAW = ("d8eb97b11b6cacbc3285473f08004500019ceccf40004006663fc0a80a7c3ff" +
+    default_key_init = SSHScan::KeyExchangeInit.new()
+    DEFAULT_KEY_INIT_RAW  = default_key_init.set_default_key_init(
+                            ("d8eb97b11b6cacbc3285473f08004500019ceccf40004006663fc0a80a7c3ff" +
                             "5db33cdfd0016982e6062988da97e801810154d2b00000101080a03a6399f3d" +
                             "f735d6000001640414e33f813f8cdcc6b00a3d852ec1aea4980000001a64696" +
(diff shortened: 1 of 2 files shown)


lib/ssh_scan/constants.rb
```diff
@@ -18,6 +21,6 @@ module Constants
                             "2d63747200000021686d61632d6d64352c686d61632d736861312c686d61632" +
                             "d726970656d6431363000000021686d61632d6d64352c686d61632d73686131" +
                             "2c686d61632d726970656d64313630000000046e6f6e65000000046e6f6e650" +
-                            "00000000000000000000000006e05b3b4").unhexify
+                            "00000000000000000000000006e05b3b4").unhexify)
```
claudijd: This isn't exactly what I was going for here, although this is a step in the right direction.

You could also achieve the same result with KeyExchangeInit because it's a bindata object class type....

default_key_init = SSHScan::KeyExchangeInit.read("...binary...data...")

What I was striving for on this was to build the object piece by piece attribute by attribute...

    default_key_init = SSHScan::KeyExchangeInit.new()
    default_key_init.macs = blah
    default_key_init.encryption = blah
    ....

claudijd: The reasoning here was mainly to be more transparent about what we're actually sending so if someone runs into a bug or compatability thing they can tell us something like "... hey I noticed you only include these ciphers, maybe you should add XYZ?"  this I think is better.

jinankjain:
@claudijd I am trying to print all those properties which we need to set, but it is kind of printing some garbage value so I am not able to get the values of all the properties

claudijd:
@jinankjain we may need to add some writer functionality to the class so it can be stood up manually.  I good place to start would be to do so here in the unit-tests => https://github.com/mozilla/ssh_scan/blob/master/spec/ssh_scan/protocol_spec.rb

jinankjain:
@claudijd We can have some setter function in that class for setting these parameters manually

claudijd:
@jinankjain yeah, that's the plan.  Bindata allows some of this to be set manually by default, but fields like length/etc. may need to be adapted as dynamic fields while building by using lambda functions.

claudijd:
@jinankjain would you like me to take a swing at this?  I have some time now to take a peek.

jinankjain:
@claudijd Yeah I am kind of busy this week but some deadlines. It would be good if you could finish this before me.

PR #113, opened 2016-08-22
Add support for custom ports in input file and support rescan

royalharsh:
```bash
```
Should fix https://github.com/mozilla/ssh_scan/issues/100 because now you can specify custom ports in the file. Am I correct @claudijd ?

Now we can use 
```bash
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,7 @@ build/
 
 # for a library or gem, you might want to ignore these files since the code is
 # intended to run in multiple environments; otherwise, check them in:
-# Gemfile.lock
+Gemfile.lock
 # .ruby-version
 # .ruby-gemset
 
--- a/bin/ssh_scan
+++ b/bin/ssh_scan
@@ -6,11 +6,13 @@ $:.unshift File.join(File.dirname(__FILE__), "../lib")
 require 'ssh_scan'
 require 'optparse'
 require 'netaddr'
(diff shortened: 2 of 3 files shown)


claudijd:
@royalharsh thanks for your contributions here, this is great.  One thing I think would be good to take care of on this pass is to address the ambiguity that's arising between the concept of a target and a port.  I wonder if a target should actually be considered a socket pairing  for example....

**Current Target**
192.168.1.1

**Current Port***
22

** Proposed Target (port is included)**

    class Target
      def initialize(ip, port = 22)
        ....  
      end
    end

This might allow us to reduce some of the complexity within /bin/ssh_scan and pull it into the library classes so it's something that TargetParser emits so we can provide stronger test coverage for it, have clear expectations on what happens in what scenarios, and less chance for bugs.

Is that something you'd be interested in working on?  If not, just let me know and I'll take what you've got here and adapt it to that strategy.  If so, just say the words and I'll wait for a PR with the changes.

royalharsh:
I think you're right, it would allow us to reduce the complexity in /bin/ssh_scan. I'd be interested in working on it! :smile:

claudijd:
@royalharsh sounds good, please make sure you rebase or start with a fresh branch from master, a lot of that code is changing rapidly day to day with lots of active contribution.  Thanks!

PR #120, opened 2016-08-23
Fixes #115

### LINKED ISSUE #115: Set a default max thread pool and expose the ability for users to control it
Currently, if we're given 100,000 IPs to scan, we'll continue to thread those out until we run out of memory.  We should have some sane limit on how big that thread pool should get at any given time for not only preservation of the scanner, but also preservation of the network we're scanning.

Recommended default threads should be something under 10.

jinankjain:
@claudijd Fixed that thread pooling thingy


claudijd:
I just fixed the same thing in #121 

Sorry for the collision, but looks like we had the same approach anyways.

PR #124, opened 2016-08-26
support output as input for rescans

### LINKED ISSUE #101: Support output as input for re-scans

agaurav77:
Fixes #101.

I have disabled support for multiple ports, since they are not supported right now. So, if an input file has two hosts
192.168.1.1, port 22
192.168.1.2, port 9999

This is treated as an invalid input file (for now). (If you want I can try adding support for different ports on different hosts?)

Thanks!

### DIFF
--- a/bin/ssh_scan
+++ b/bin/ssh_scan
@@ -6,6 +6,7 @@ $:.unshift File.join(File.dirname(__FILE__), "../lib")
 require 'ssh_scan'
 require 'optparse'
 require 'netaddr'
+require 'json'
 
 #Default options
 options = {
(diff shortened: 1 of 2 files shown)


claudijd:
@agaurav77 thanks for you contribution here.  @royalharsh submit #125 shortly after that took a similar strategy and included some other features and tests, so I'm closing this because it's a subset of #125.

agaurav77:
@claudijd No worries, glad to know :)

PR #125, opened 2016-08-28
Add support for custom ports in input file and support rescans

royalharsh:
```bash
```

Now we can use 
```bash
```
How about this? @claudijd

### DIFF
--- a/bin/ssh_scan
+++ b/bin/ssh_scan
@@ -3,14 +3,14 @@
 # Path setting slight of hand
 $:.unshift File.join(File.dirname(__FILE__), "../lib")
 
-require 'ssh_scan'
-require 'optparse'
+require 'json'
 require 'netaddr'
+require 'optparse'
+require 'ssh_scan'
 
 #Default options
 options = {
-  :targets => [],
-  :port => 22,
+  :sockets => [],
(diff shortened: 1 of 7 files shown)


claudijd:
@royalharsh many thanks for taking this on.  I think this is a good step for a couple feature we want to improve on.  Also, thank you for the tests!

bin/ssh_scan
```diff
@@ -48,14 +50,30 @@ opt_parser = OptionParser.new do |opts|
...
   opts.on("-o", "--output [FilePath]",
           "File to write JSON output to") do |file|
     $stdout.reopen(file, "w")
   end
 
   opts.on("-p", "--port [PORT]",
           "Port (Default: 22)") do |port|
-    options[:port] = port.to_i
+    socket = options[:sockets].shift
+    ip = socket.chomp.split(':').shift
```
jinankjain: @royalharsh @claudijd I have a doubt here with -p as an argument we directly pass port we don't have ip.

claudijd: Since this is a PR and I've already merged this code, I'll to create an issue to track it as a bug until it's resolved.

PR #129, opened 2016-08-29
Make SSHScan::KeyExchangeInit buildable from scratch

### LINKED ISSUE #104: Define default client key_init message as an object rather than raw blob

claudijd:
Fixes #104

### DIFF
--- a/lib/ssh_scan/client.rb
+++ b/lib/ssh_scan/client.rb
@@ -19,7 +19,7 @@ module SSHScan
       @port = port
       @client_banner = SSHScan::Constants::DEFAULT_CLIENT_BANNER
       @server_banner = nil
-      @kex_init_raw = SSHScan::Constants::DEFAULT_KEY_INIT_RAW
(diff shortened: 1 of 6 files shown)


claudijd:
@jinankjain this ended up turning out to be way more involved than I had originally expected.  Alas, it's sorted, but might benefit from some refactoring/simplification down the road

claudijd:
Also, on the plus side, found and fixed a couple bugs in the protocol while I was in there.

PR #132, opened 2016-08-29
Fixes #74

### LINKED ISSUE #74: Add support for IPv4 fallback when IPv6 cannot be established
Basically, if we have an endpoint that supports both IPv6 and IPv4 on the same hostname (ssh.example.com) and that endpoint has records for both IPv6 and IPv4, but only IPv4 has a listening SSH service we will likely not handle that scenario.  This issue is to make sure ssh_scan does have the smarts to handle that gracefully.

jinankjain:
@claudijd What do you say about this one?

### DIFF
--- a/lib/ssh_scan/scan_engine.rb
+++ b/lib/ssh_scan/scan_engine.rb
@@ -9,11 +9,34 @@ module SSHScan
       target, port = socket.chomp.split(':')
       policy = opts[:policy_file]
       timeout = opts[:timeout]
+      result = []
 
-      client = SSHScan::Client.new(target, port, timeout)
-      client.connect()
-      result = client.get_kex_result()
-      return result if result.include?(:error)
+      if target.fqdn?
+        if target.resolve_fqdn_as_ipv6.nil?
+          client = SSHScan::Client.new(target.resolve_fqdn_as_ipv4.to_s, port, timeout)
+          client.connect()
+          result = client.get_kex_result()
+          result[:hostname] = target
+          return result if result.include?(:error)
+        else
+          client = SSHScan::Client.new(target.resolve_fqdn_as_ipv6.to_s, port, timeout)
+          client.connect()
+          result = client.get_kex_result()
+          if result.include?(:error)
+            client = SSHScan::Client.new(target.resolve_fqdn_as_ipv4.to_s, port, timeout)
+            client.connect()
+            result = client.get_kex_result()
+            result[:hostname] = target
+            return result if result.include?(:error)
+          end
+        end
+      else
+        client = SSHScan::Client.new(target, port, timeout)
+        client.connect()
+        result = client.get_kex_result()
+        result[:hostname] = ""
+        return result if result.include?(:error)
+      end
 
       # Connect and get results (Net-SSH)
       begin
--- a/lib/string_ext.rb
+++ b/lib/string_ext.rb
@@ -1,4 +1,5 @@
 require 'ipaddr'
+require 'resolv'
 
 # Extend string to include some helpful stuff
 class String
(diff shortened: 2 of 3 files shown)


lib/string_ext.rb
```diff
@@ -22,6 +23,22 @@ def ip_addr?
...
+    Resolv::DNS.open do |dns|
+      ress = dns.getresources self, Resolv::DNS::Resource::IN::AAAA
+      temp = ress.map { |r| r.address  }
+      return temp[0]
+    end
+  end
+
+  def resolve_fqdn_as_ipv4
+    Resolv::DNS.open do |dns|
+      ress = dns.getresources "github.com", Resolv::DNS::Resource::IN::A
```
claudijd: You have a hardcoded value here.  Needs to be fixed before landing.

lib/ssh_scan/scan_engine.rb
```diff
@@ -9,11 +9,34 @@ def scan_target(socket, opts)
...
-      return result if result.include?(:error)
+      if target.fqdn?
+        if target.resolve_fqdn_as_ipv6.nil?
+          client = SSHScan::Client.new(target.resolve_fqdn_as_ipv4.to_s, port, timeout)
+          client.connect()
+          result = client.get_kex_result()
+          result[:hostname] = target
+          return result if result.include?(:error)
+        else
+          client = SSHScan::Client.new(target.resolve_fqdn_as_ipv6.to_s, port, timeout)
```
claudijd: What is the expected behavior when a host is not dual-stacked (v4 only), but the DNS entry for a AAAA record for say github.com has a valid IPv6 address?

jinankjain:
@claudijd Updated

lib/ssh_scan/scan_engine.rb
claudijd: I think we might need some localized code to determine if the scanner itself can route v6 traffic to the endpoint and maybe a more explicit error that says "Hey, I can only scan v4 hosts" or something.

jinankjain: @claudijd I did not get you still.

claudijd: @jinankjain from the perspective of the scanner, in order for it to talk to IPv4 and IPv6 targets, the scanner itself will need to support IPv4 and IPv6 networking.  It's pretty common these days to see most hosts support IPv4, but not IPv6.  However, our DNS queries for "github.com" for the A and AAAA DNS records will happen over whatever the default is for that host.  So if the host does not have IPv6 networking, but it can still do a DNS lookup for a AAAA record and get an IPv6 address what we expect to happen?

jinankjain: @claudijd That is what exactly happening now I am first looking for AAAA records and if I get nil value, then I look for A records for IPv4 and in the last if I handle the case when IPv6 does not support connection to port then I go for IPv4 connection

claudijd: @jinankjain ok, I missed that.

lib/string_ext.rb
```diff
@@ -22,6 +23,22 @@ def ip_addr?
     return true
   end
 
+  def resolve_fqdn_as_ipv6
+    Resolv::DNS.open do |dns|
+      ress = dns.getresources self, Resolv::DNS::Resource::IN::AAAA
+      temp = ress.map { |r| r.address  }
```
jinankjain: @claudijd This is the place where I am resolving from AAAA

lib/string_ext.rb
```diff
@@ -22,6 +23,22 @@ def ip_addr?
...
+      return temp[0]
+    end
+  end
+
+  def resolve_fqdn_as_ipv4
+    Resolv::DNS.open do |dns|
+      ress = dns.getresources self, Resolv::DNS::Resource::IN::A
+      temp = ress.map { |r| r.address  }
```
jinankjain: @claudijd  This is the place where I am resolving from A

PR #142, opened 2016-09-01
Added auth_methods as a policy control. Fixes #139

### LINKED ISSUE #139: Add auth_methods as a policy control
It would be nice if we added auth_methods as a policy control.

For example:

If someone wants to say "though shall only use public_key auth" we should be able to make that a compliance item and false compliance because of it.

yashmehrotra:
To add a whitelist of auth_methods, we just need to add the below sample block to the `policy.yml` file:

```yml
auth_methods:
- publickey
- password
```

If no whitelist of auth_methods is provided, it assumes all `auth_methods` are valid.

### DIFF
--- a/lib/ssh_scan/policy.rb
+++ b/lib/ssh_scan/policy.rb
@@ -2,7 +2,7 @@ require 'yaml'
 
 module SSHScan
   class Policy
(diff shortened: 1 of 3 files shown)


claudijd:
nice!  double thank you for including unit-tests!

Issue #140, opened 2016-09-01
resolv.rb:164:in `<class:Resolv>': uninitialized constant Timeout::Error (NameError)


christophprokop:
Hi there,

i compiled (core) ruby and then installed ssh_scan via
`$ gem install ssh_scan`
It works just fine with version 0.0.9 and 0.0.10, but when i try to execute ssh_scan in version 0.0.11 (e.g. with -h flag) i get:

```
$ /usr/local/ruby/2.3.1/bin/ssh_scan -h
/usr/local/ruby/2.3.1/lib/ruby/2.3.0/resolv.rb:164:in `<class:Resolv>': uninitialized constant Timeout::Error (NameError)
Did you mean?  KeyError
               IOError
               Errno
        from /usr/local/ruby/2.3.1/lib/ruby/2.3.0/resolv.rb:39:in `<top (required)>'
        from /usr/local/ruby/2.3.1/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
        from /usr/local/ruby/2.3.1/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
        from /usr/local/ruby/2.3.1/lib/ruby/gems/2.3.0/gems/ssh_scan-0.0.11/lib/string_ext.rb:2:in `<top (required)>'
        from /usr/local/ruby/2.3.1/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
        from /usr/local/ruby/2.3.1/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
        from /usr/local/ruby/2.3.1/lib/ruby/gems/2.3.0/gems/ssh_scan-0.0.11/lib/ssh_scan/constants.rb:1:in `<top (required)>'
        from /usr/local/ruby/2.3.1/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
        from /usr/local/ruby/2.3.1/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
        from /usr/local/ruby/2.3.1/lib/ruby/gems/2.3.0/gems/ssh_scan-0.0.11/lib/ssh_scan.rb:6:in `<top (required)>'
        from /usr/local/ruby/2.3.1/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
[...]

PR #145, opened 2016-09-01
Add gem-based integration tests

claudijd:


claudijd:
There two failures...

https://travis-ci.org/mozilla/ssh_scan/jobs/156910905 (not expected - need to figure out why docker matric is inheriting integration test runs)
https://travis-ci.org/mozilla/ssh_scan/jobs/156910902 (expected because of bug #140)

claudijd:
We are now failed as we would expect, no surprises now.

Issue #150, opened 2016-09-02
Add ssh_lib fingerprint for dropbear ssh


claudijd:

jinankjain:
Closed via #165

PR #165, opened 2016-09-05
add dropbear ssh_lib fingerprints

agaurav77:
For #150. Also added banner tests for the same.


lib/ssh_scan/ssh_lib/dropbear.rb
```diff
@@ -0,0 +1,55 @@
...
+      @@VERSIONS = ["2016.74", "2016.73", "2016.72", "2015.71", "2015.70",
+                    "2015.69", "2015.68", "2015.67", "2014.66", "2014.65",
+                    "2014.64", "2014.63", "2013.62", "2013.61test", "2013.60",
+                    "2013.59", "2013.58", "2013.57", "2013.56", "2012.55",
+                    "2011.54", "0.53.1", "0.53", "0.52", "0.51", "0.50",
+                    "0.49", "0.48.1", "0.48", "0.47", "0.46", "0.45", "0.44",
+                    "0.44test4", "0.44test3", "0.44test2", "0.44test1", "0.43",
+                    "0.42", "0.41", "0.40", "0.39", "0.38", "0.37", "0.36",
+                    "0.35", "0.34", "0.33", "0.32", "0.31", "0.30", "0.29",
+                    "0.28"]
```
claudijd: @agaurav77 what purpose does the version table have in this implementation?  Could this just as easily be included in unit-tests and have a generic reg-ex or set of regexes detect these?  I ask because this because next time a new version is released we have to update our version guesser before that version will be recognized.  I think it would be better if the version class would just support the various formats seen with this code...

Here's a guess at a format that might support them all, but it would be good to have unit-tests that cover those uniq examples to make sure....

    \d+(\.\d+)+(test\d+)?

claudijd:
Added a comment about the versions array and suggested an alternative.  This also will need a rebase.

claudijd:
Also created this to eliminate future merge conflicts on fingerprint updates => #168

agaurav77:
@claudijd You are right, this can very easily be matched against a simple regex. I did something similar in one of the files (/SSH-2.0-dropbear_(\d+\.\d+(?:test\d)?)/). I just have this array since I scraped off the versions from the Changelog and I thought it would be authentic if we matched the versions before saying they were actual versions.

For eg, 2014.56 would match against this regex, and it looks authentic, but it is not an actual version. What do you think we should do for this, then?

At best, this is a guess, so I'm hoping it'd be okay if we don't keep the versions array (since we'll have to update things after a while), and instead guess on a version. But are we willing to compromise on the correctness of the reported version? Just a genuine doubt.

claudijd:
@agaurav77 I think it's fine to trust the system to report whatever version it is and to parse the various format examples we have.  If tomorrow they release 2016.99 we will not be able to fingerprint it, and that to me makes this fingerprinting too rigid.  I'd rather just trust the endpoint to report it's version.

agaurav77:
@claudijd I added the VERSIONS array in the tests, and updated the regex to match all strings (and hopefully further version strings). Please have a look, thanks!

claudijd:
landed, thanks! This should free up the effort to finish up the fingerprint breakout effort.

Issue #130, opened 2016-08-29
Potential port bug found in already merged #125


claudijd:
https://github.com/mozilla/ssh_scan/pull/125

https://github.com/mozilla/ssh_scan/commit/18e958d7f9ca9347cf0bda2ba7ab304dc84de592#diff-903158d449f10b1aac623831242f53e1R41

@jinankjain said...

"@royalharsh @claudijd I have a doubt here with -p as an argument we directly pass port that would applied to all the passed target not just the last one."

royalharsh:
Does this mean while doing this
```bash
./bin/ssh_scan -t ssh.mozilla.com,github.com -p 2222
```
We want port as 2222 for only github.com and not for ssh.mozilla.com? @claudijd

jinankjain:
We want for both @royalharsh

claudijd:
This is probably a good time to call out what our expectations should be for how ports are applied in various scenarios so we can ensure we have tests that verify these cases....

**Easy Scenarios**
1. User does -t scan with no -p port specification
- all targets via -t will be scanned on port 22
2.) User does -t scan with -p port specification
- all targets via -t will be scanned via -p port specification, example 22222
3.) User does -t scan with -p multiple port specification
- all targets via -t will be scanned via all ports in -p port specification, example 22,22222
4.) User does -f scan with no -p port specification
- all targets via -f file will be scanned on port 22
5.) User does -f scan with -p port specification
- all targets via -f file will be scanned via -p port specification, example 22222

**Harder Scenarios**
1. User does -f scan with -p port specification and targets in -f also contain :port specifications
- ??? I think -f :port specification should override -p specification
2.) User does -t scan and -f scan at the same time
- ??? should throw an exception making options mutually exclusive
3.) User does -t scan with socket inputs (example: github.com:22)
- ??? I dunno
...INSERT_MORE - we can discuss and then implement specs that ensure it behaves the way we want to.

jinankjain:
@claudijd For harder scenarios I think there is a way out

-  User does -f scan with -p port specification and targets in -f also contain :port specifications. Instead of overriding the ports we could create a set which do union of both the port specification. Same applies for -t and -f together

claudijd:
Another idea would be to consider the following...

-t  and -f flags triggers an exception (can't use them together)
-p and socket-based targets (via -t or -f) triggers an exception (can't use them together)

This could vastly simplify the above, but either way I think it's possible.  Just need feedback on what makes the most sense.

[...]

PR #166, opened 2016-09-05
fix nit in #130

agaurav77:
@claudijd 

I tried ssh_scan -t ip -p 9999, and upon inspecting options[:sockets] it looked something like ["ip:22", "ip:9999"], when it should've been ["ip:9999"].

### DIFF
--- a/lib/ssh_scan/target_parser.rb
+++ b/lib/ssh_scan/target_parser.rb
@@ -1,12 +1,13 @@
 require 'netaddr'
 require 'string_ext'
+require 'set'
 
 module SSHScan
   class TargetParser
-    def enumerateIPRange(ip, port = "22")
+    def enumerateIPRange(ip, port = "22", port_append = false)
       if ip.fqdn?
-        socket = ip.concat(":").concat(port.to_s)
-        return [socket]
+        socket = port_append ? ip.concat(":").concat(port.to_s) : ip
+        return [socket].to_set
       else
         if ip.include? "-"
           octets = ip.split('.')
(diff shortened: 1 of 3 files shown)


jinankjain:
I made it like that but I think we need to decide what should be the final call for this case -p should overwrite the existing port or not?

jinankjain:
@agaurav77 Your fix will overwrite this github.com:33 -p 333,3333 to github.com:333 and github.com:3333

agaurav77:
@jinankjain You're right, but there is a side case as well. When the user has specified a port via -p and not specified it via ip:port, then I think he expects the port provided via -p to override the default value (22).

agaurav77:
I think we should append extra ports specified via -p (as it is right now), but also handle the case when ssh_scan makes ip = ip:22 when the user didn't specify the port via the colon notation, and specified it instead via -p. So, this might need more change.

jinankjain:
We should handle this case too github.com:22 -p 33,333

agaurav77:
@jinankjain How about this?

This gives
1) github.com:22 -p 33,333
github.com:22
github.com:33
github.com:333

2) github.com -p 33
github.com:33

3) github.com
github.com:22

jinankjain:
@agaurav77 We could keep all these sockets in a set instead of array so that we don't have to worry about duplicate sockets entries.

agaurav77:
@jinankjain Good point, I'll use Set for the IP Range then :)

agaurav77:
@jinankjain Done :)

lib/ssh_scan/target_parser.rb
```diff
@@ -1,33 +1,50 @@
...
+          socket = port_append ? ip.concat(":").concat(port.to_s) : ip
+          return [socket].to_set
         end
       end
     end
+
+    # Remove instances of ip, where ip:port already exist.
+    # This would mean that port was somehow passed. Otherwise,
+    # make ip = ip:22
+    def sanitizeIPRange(ip_range)
```
jinankjain: @agaurav77 Do we need this after introducing set?

agaurav77: @jinankjain yes

Maybe I got you confused. Let me clarify once more.

Right now, when you pass github.com -p 99, it checks github.com:22, github.com:99, which is a side effect of appending ports in -p to the targets given in -t. Since the default port is 22, the first time you were doing enumerateIPRange, it would convert github.com to github.com:22, and then simply append 99 to the host github.com again. But then this 22 is kind of redundant.

The need for my logic stems from the fact that there is (maybe) no way to check whether -p was passed within -t block in OptionParser's code. Thus, enumerate in my PR converts github.com -p 99 to github.com, github.com:99, and then sanitizes it to remove github.com since a github.com:port argument is present. Maybe there are shorter ways to do this, and please feel free to indicate them, I'd be more than happy to change my logic to something shorter.

Sorry for the long explanation :)

claudijd:
@jinankjain @agaurav77 glad you guys are tracking this down.

I think the easiest way to handle this could be to make socket specifications via -t incompatible with -p to make it easier.  I think the scenarios are getting a little unwieldy and it becomes hard at a point for a new user to understand the implications of joining socket specifications.  The other option, which seems to make sense to me is that when you specify a target with an explicit socket specification, it would fully override the port specification of that target.

I also agree that a -p 22222 should not result in 22 and 22222.   If you choose -p 22222 it should only scan 22222.  If you want to scan both, you should have to specify -p 22,22222.

In the end, I'm really happy that we went the target parser route, I think this helps make expectations more clear.  Perhaps what we should do is make a direct mapping from command-line switches to target parser options so we don't loose any context when target parser goes to interpret?  Then we can easily spec in the behavior and if we ever want to change it we can do so explicitly.

agaurav77:
@claudijd So then do you want something like (in pseudocode) this?

Parse all passed command line switches
if -t and no -p then ...
if -t and -p then ...
...

I'm sorry if I misunderstood your idea.

claudijd:
@agaurav77 some like that...

- if -t github.com:22 -p 2222 then exception
- if -f <with github.com:22> -p 2222 then exception

jinankjain:
@agaurav77 I had fixed this up in #171

claudijd:
#171 is now landed, closing this for now, please reopen if needed.

PR #171, opened 2016-09-06
Fixed port issue

jinankjain:
@claudijd  @agaurav77  Fixed that port issue by overwriting ports with what is specified in -p

### DIFF
--- a/bin/ssh_scan
+++ b/bin/ssh_scan
@@ -27,9 +27,10 @@ opt_parser = OptionParser.new do |opts|
                 "Usage: ssh_scan [options]"
 
   opts.on("-t", "--target [IP/Range/Hostname]", Array,
-          "IP/Ranges/Hostname to scan") do |ips|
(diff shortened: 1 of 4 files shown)


jinankjain:
For more clarity 

-t github.com:22 -p 33,3333 final output github.com:33, github.com:3333
Same is true for -f flag too

This solves all the confusion

claudijd:
@jinankjain if we're going to bury the :22 input, I don't think it should be silent.  I think this should throw an error.

jinankjain:
@claudijd So we throw an error message stating that port 22 is rejected?

claudijd:
@jinankjain I was thinking more a generic hard failure... -t flag with socket target + -p flags are incompatible

jinankjain:
Okay cool that's kind of easy to do. I will update the patch soon

jinankjain:
@claudijd Updated the patch

Issue #168, opened 2016-09-06
Break fingerprint specs into their own respective files


claudijd:
When we have multiple people contributing fingerprints this is aweome, but when we have multiple concurrent adds to the single banner specs file it results in merge conflicts requiring contributors to have to rebase before their add is included.  We want to get rid of this two-step process so people can send us whatever they want and have it stay mergable unless someone is editing the same fingerprint class in which case a conflict makes sense.

rishabhs95:
Yeah, I think it will be much more convenient for contributors as well... will make the specs more modular :)

agaurav77:
@claudijd Should we close this now? Or something else needed?

claudijd:
Nope, it's done, thanks!

rishabhs95:
great work @agaurav77! really liked the new modularised code structure :)

PR #173, opened 2016-09-07
break fingerprint specs into their own files

agaurav77:
For #168 

This way, anyone who wants to write a spec test for an OS (say Debian) can make a file in spec/ssh_scan as os/debian_spec.rb

```
require 'rspec'
require_relative '../helper'

describe ...
  context ...
    expectations = {
      .. fingerprints ..
    }
    checkFingerprints(expectations)
  end
end
```

I hope this make it easier. Please do suggest ways to improve this structure of tests, thanks!

### DIFF
--- a/spec/ssh_scan/banner/banner_spec.rb
+++ b/spec/ssh_scan/banner/banner_spec.rb
@@ -1,0 +1,13 @@
+require 'rspec'
+require_relative 'helper'
+
+describe SSHScan::Banner do
+  context "when creating from scratch" do
+    it "it should create a new Banner object" do
+      banner_string = "SSH-2.0-server"
+      banner = SSHScan::Banner.new(banner_string)
+      expect(banner).to be_kind_of(SSHScan::Banner)
+      expect(banner.to_s).to eql(banner_string)
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


agaurav77:
I would've added the ssh_lib folder the same way for specific ssh_lib tests, say ssh_lib/dropbear_spec.rb (that PR is not submitted yet, so not present in this PR)

claudijd:
@rishabhs95 was also looking to work on this, CC'ing just as an FYI

claudijd:
This branch now needs a rebase to make it ready after the drop bear PR was merged into master.

agaurav77:
@claudijd Sure, I'll rebase this.

claudijd:
@agaurav77 hopefully after we get this all refactored it will eliminate a lot of the rebasing problems we've had recently.

agaurav77:
@claudijd I rebased the PR. Putting that aside, does the modularization seem okay? Any suggestions?

claudijd:
@agaurav77 I think this is really good.  I really like the fact you factored out the core logic like you did, helps keep things more DRY (Don't Repeat Yourself).  I think could run into rebase conflicts on spec/ssh_scan/banner/generic_spec.rb, but I think those cases will be rare enough to not worry about it now.  If for some reason we're constantly adding to it, then maybe we'll need to specialize more, but this is good for now.  Will land in a sec.

PR #176, opened 2016-09-07
Added SSH version as a policy control. Fixes #154

### LINKED ISSUE #154: Add SSH version as a policy control
Perhaps we want to require that SSH be version 2.0

Maybe this would be a bad example?

 "ssh_version": 1.99,

yashmehrotra:
We just need to add `ssh_version: <minimum>` to the `yml` file.

### DIFF
--- a/lib/ssh_scan/policy_manager.rb
+++ b/lib/ssh_scan/policy_manager.rb
@@ -104,6 +104,16 @@ module SSHScan
       return outliers
     end
 
+    def out_of_policy_ssh_version
+      target_ssh_version = @result[:ssh_version]
+      if @policy.ssh_version
+        if target_ssh_version < @policy.ssh_version
+          return true
+        end
+      end
+      return false
+    end
+
     def compliant?
       out_of_policy_encryption.empty? &&
       out_of_policy_macs.empty? &&
@@ -113,7 +123,8 @@ module SSHScan
       missing_policy_macs.empty? &&
       missing_policy_kex.empty? &&
       missing_policy_compression.empty? &&
-      out_of_policy_auth_methods.empty?
+      out_of_policy_auth_methods.empty? &&
+      out_of_policy_ssh_version
     end
 
     def recommendations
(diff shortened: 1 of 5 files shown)


claudijd:
@yashmehrotra please feel free to add 2.0 to modern and intermediate policies here in this PR => https://github.com/mozilla/ssh_scan/tree/master/policies

claudijd:
@yashmehrotra also, feel free to join us in Glitter chat if you'd like.  I'd invite you directly, but don't have your email (https://gitter.im/mozilla-ssh_scan/Lobby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)

lib/ssh_scan/policy_manager.rb
```diff
@@ -104,6 +104,16 @@ def out_of_policy_auth_methods
...
 
+    def minimum_ssh_version
+      target_ssh_version = @result[:ssh_version]
+      if @policy.ssh_version
+        if target_ssh_version < @policy.ssh_version
+          return @policy.ssh_version
+        end
+      end
+      return false
+    end
```
claudijd: Returning false or string is a bit weird here.

Maybe we need to improve terminology here.

Maybe there is a method that returns true/false, one that returns the version we saw, and one that returns the version we want?  I feel like that might be a little more deliberate.

spec/ssh_scan/policy_manager_spec.rb
```diff
@@ -192,4 +192,25 @@
...
+                  "auth_methods:\n- publickey\n" +
+                  "ssh_version: 2.0\n"
+    result = {
+                :ssh_version => 1.9
+             }
+
+    it "should load all the attributes properly" do
+      policy = SSHScan::Policy.from_string(yaml_string)
+      policy_manager = SSHScan::PolicyManager.new(result, policy)
+      expect(policy_manager.minimum_ssh_version()).to eql(2.0)
```
claudijd: Same vein as the method criticism, I think we need more deliberate methods here.  We're using one method for multiple uses and I don't think it's direct enough.

yashmehrotra: @claudijd So what do you suggest for this ?

claudijd: very the response of "policy_mgr.compliance_results" that you expect

yashmehrotra:
@claudijd Hi, I used a new approach, where the function that checks the minimum ssh version only returns a boolean. I hope its good.

claudijd:
@yashmehrotra thanks!

Issue #84, opened 2016-08-15
Abstract host key fingerprint generation out of scan_engine into it's own thing


claudijd:
Here's some basic guidelines...

1.) Must support RSA, DSA, and ECDSA
2.) Must generate md5, sha1, and sha256 fingeprints (with an eye toward other fingerprint styles in use or upcoming)
3.) Must be fully spec'd out so we easily prove in unit-tests that our coverage is good and works as expected (perhaps even using openssh/openssl as a verification point before we lock in the fingerprints in the specs)
4.) Must have a soft-failure mechanism when the host key type is unsupported and then send a warning to the command-line much like the current implementation does

flash1452:
@claudijd I would like to work on this issue.

claudijd:
@flash1452 sounds good, give it a whirl.  Thanks!

PR #185, opened 2016-09-12
take out host key fingerprinting from scan_engine

agaurav77:
For #84.

I had some doubts on how to generate the ECDSA fingerprints, so didn't push that change in this PR.

### DIFF
--- a/lib/ssh_scan/crypto.rb
+++ b/lib/ssh_scan/crypto.rb
@@ -1,0 +1,38 @@
+require 'openssl'
+
+module SSHScan
+  module Crypto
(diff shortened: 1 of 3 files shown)


agaurav77:
@claudijd Please review

claudijd:
@agaurav77 nice job with this, I think we should create some unit-tests around this newly created abstraction.  This is something we weren't able to do before because it wasn't abstracted out.

agaurav77:
@claudijd Should I add more examples?

claudijd:
@agaurav77 nice job here!

PR #194, opened 2016-09-19
SQLite prototype shim example

claudijd:
Example output modification, when we see an SSH endpoint where the host_key is re-used and the implication becomes that there is a private key dependency with these other hosts....

    "fingerprints": {
      "md5": "1a:8f:af:7d:4d:f7:2e:9d:e7:69:bb:07:17:4c:b0:71",
      "sha1": "b1:0c:5c:58:09:eb:20:3f:10:bc:e1:8b:0e:d8:fd:55:a8:21:60:d4",
      "sha256": "8a:92:84:06:ca:08:64:85:28:a0:36:0f:82:f3:dc:75:01:fc:01:82:35:e8:93:28:9d:e3:aa:b0:33:59:61:16"
    },
    "duplicate_fingerprints": [
      "192.168.10.100"
    ],

This new scan result basically tells us that .99 (the host we scanned) and the .100 host are sharing host_keys, which would be bad in a private key disclosure scenario.

### DIFF
--- a/lib/ssh_scan/fingerprint_database.rb
+++ b/lib/ssh_scan/fingerprint_database.rb
@@ -1,0 +1,39 @@
+require 'sqlite3'
+
+module SSHScan
+  class FingerprintDatabase
+    def initialize(database_name)
+      if File.exists?(database_name)
+        @db = ::SQLite3::Database.open(database_name)
+      else
+        @db = ::SQLite3::Database.new(database_name)
+        self.create_schema
+      end
+    end
+
+    def create_schema
+      @db.execute <<-SQL
+        create table fingerprints (
+          fingerprint varchar(100),
+          ip varchar(100)
+        );
+      SQL
+    end
+
+    def clear_fingerprints(ip)
+      @db.execute "delete from fingerprints where ip like ( ? )", [ip]
+    end
+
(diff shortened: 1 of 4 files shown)


claudijd:
Known gaps include the fact that duplicates are processed only at the scan time of a given host (ie. no post processing  in this example).  What this means is that if host A and host B are sharing keys, and we scan A first then B second that only B will report duplicates.  I think this means that we'll need to re architect duplicate host keys to be a post-process sort of decorator, and if we want to have compliance status depend on this, compliance status will need to do the same.

claudijd:
Ok, yeah, so fingerprint duplicates and compliance status should be decorators for the sake of accommodating all consumers (web/api/command-line).

It should work like this....

1.) Scan all hosts and retrieve attributes about the service
2.) Cross reference all hosts against either a local fingerprint database or an api database (depending on implementation) for duplicates and add duplicate decorators
3.) Lastly, apply compliance decorators

This will afford use a lot of flexibility in the long run if we consider the ability to apply a compliance filter over a static set of SSH results.  This would allow people to experiment with different policy layers without needing to re-scan, which is highly lucrative.

claudijd:
Superceded by #197

PR #197, opened 2016-09-19
Add persistent fingerprint DB capability

claudijd:
To reproduce this behavior, it requires a little hacking if you don't have an example target...

You're need to scan a host to generate a local fingerprint DB stub

./bin/ssh_scan -t 192.168.10.99

At this time, you shouldn't have any duplicates in your results.  To insert a duplicate into the db you can do this...

sqlite3 fingerprints.db
insert into fingerprints values ('8a:92:84:06:ca:08:64:85:28:a0:36:0f:82:f3:dc:75:01:fc:01:82:35:e8:93:28:9d:e3:aa:b0:33:59:61:16', '192.168.10.100');

Make sure to use a fingerprint string from the original scan.

Once this is added, you can now rescan and get the following host key duplicate decorators...

    "duplicate_host_key_ips": [
      "192.168.10.100"
    ],

This new information added to scans will let us know if we ever come across a host that is sharing keys with another host.  This means we compromise one host, we'll have the private keys to decode traffic from the other.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.gem
 *.rbc
+*.db
 /.config
 /coverage/
 /InstalledFiles
--- a/bin/ssh_scan
+++ b/bin/ssh_scan
@@ -18,6 +18,7 @@ options = {
   :threads => 5,
   :verbosity => nil,
   :logger => Logger.new(STDERR),
+  :fingerprint_database => "./fingerprints.db"
 }
 
 target_parser = SSHScan::TargetParser.new()
--- a/bin/ssh_scan
+++ b/bin/ssh_scan
@@ -108,13 +108,18 @@ opt_parser = OptionParser.new do |opts|
     options[:threads] = threads.to_i
   end
 
+  opts.on("--fingerprint-db [FILE]",
+          "File location of fingerprint database (Default: ./fingerprints.db)") do |fingerprint_db|
+    options[:fingerprint_database] = fingerprint_db
+  end
+
   opts.on("-u", "--unit-test [FILE]",
           "Throw appropriate exit codes based on compliance status") do
     options[:unit_test] = true
   end
 
   opts.on("-V", "--verbosity",
-          "Set the logger level (Accpeted Params: INFO, DEBUG, WARN, ERROR, FATAL)") do |verbosity|
+          "Set the logger level (Accepted Params: INFO, DEBUG, WARN, ERROR, FATAL)") do |verbosity|
     options[:logger].level = case options[:verbosity]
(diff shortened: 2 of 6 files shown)


claudijd:
This also makes fingerprint storage persistent across multiple scans.

claudijd:
@jinankjain would you mind peer reviewing this and maybe even trying a scan before I land this?

jinankjain:
@claudijd Yeah sure. I will do it in the evening 😄

jinankjain:
@claudijd One thing that needs to be done is add fingerprints.db file in .gitignore so that it does not get pushed

jinankjain:
@claudijd One more thing we need to take care is if we add multiple entries of the same ip then we get all of them listed in the json data

```
duplicate_host_key_ips": [
      "192.30.253.112",
      "192.30.253.112",
      "192.30.253.112"
    ],
```

jinankjain:
@claudijd One more thing I want to ask why did you switch from mongo to sqlite3?

claudijd:
@jinankjain I switched from mongodb to sqlite because it was less setup for end users.  Differences are that for sqlite you don't need to host a service, where with mongodb you do.  I really like the nosql benefits of mongodb, but it was just too much to ask of a command-line user IMO.  However, if we have a server-side deployment, where more setup is ok, we can provide an option for mongodb, so we can store full scan result data.

claudijd:
I'll add some commits for the other comments.  Thanks for the review.

jinankjain:
@claudijd How about one more thing adding a command line parameter for mentioning database file?

claudijd:
@jinankjain yeah, will do, thanks!

claudijd:
$ ./bin/ssh_scan -t github.com --fingerprint-db test.db

$ sqlite3 test.db 
SQLite version 3.8.5 2014-08-15 22:37:57
Enter ".help" for usage hints.
sqlite> select * from fingerprints;
f6:ce:1e:d1:87:52:2d:26:81:89:ec:e6:d3:8f:9f:69|192.30.253.113
4d:b1:29:4f:e8:de:1c:14:1f:11:47:45:4e:8c:9a:c8:a8:45:76:40|192.30.253.113
9e:16:bf:1a:99:1b:b9:f6:9a:f9:24:fd:b2:55:29:51:d8:d2:20:06:14:c8:03:b0:bf:bb:95:9a:d1:a7:83:0f|192.30.253.113

claudijd:
Also, yay for integration tests (https://travis-ci.org/mozilla/ssh_scan/jobs/161346087).  Helped find that scoping bug, but unit-tests passed.

Issue #50, opened 2016-08-03
Create a pretty web-ui for ssh_scan


claudijd:
Two main reasons for this:

1.) It helps make ssh_scan more approachable for the casual user that just wants to see what it can do
2.) This could serve as a reporting mechanism for self-evaluation and global stats for ssh end point configuration

Ideally, this would also plug into a queueing/job/workload engine so jobs can be tasked but are not blocking on the web-ui interface.

jinankjain:
Are you thinking of using any framework for building the web-ui?

claudijd:
@jinankjain I don't have any plans yet.  I would love if it were in Ruby, just because that's my strongest lang, but I'm open to ideas/suggestions/prototypes.

marumari:
If you wanted to simply make a nice API, I could add support to it to the Observatory's web UI.

claudijd:
@marumari the plan is to eventually have an API for programmatic access.  I would also like a simple front end for the ssh functionality, but happy to have it integrated into HTTP obs.

marumari:
Well, it would be easy to make its own page in the Observatory, or part of the normal scan stuff.  No sense in reimplementing an entire new front end.  :)

claudijd:
@marumari that sounds cool to me, where would I got to add the page?  I'll also freely admit my web dev skills are not strong, so I'm happy to defer to others that know more than I in that space.

marumari:
https://github.com/mozilla/http-observatory-website

But if you can get the API going, I'd be happy to do the web dev part for you.

claudijd:
@marumari awesome, thanks!

jinankjain:
@claudijd For this we need to work on API structuring

claudijd:
@jinankjain agreed, might be worth just drafting a doc/wiki on how it could work and then inviting @marumari for feedback since she might be helping us integrate with Observatory at some point.

agaurav77:
@claudijd @jinankjain How about Rails/Sinatra for the backend? Rails might be a bit bulky though (convention over configuration ends up in a lot of unnecessary files). Sinatra might be just the thing for the API, though.

claudijd:
I'd vote for Sinatra over Rails (rails is too heavy for this purpose IMO)

agaurav77:
@claudijd The basic functionality could be - 

```
require 'sinatra'

get '/scan/:host' do |host|
  # scan hostname / single IP / IP Range
end

get '/scan' do
[...]

PR #198, opened 2016-09-19
add api.md for web ui api

agaurav77:
For #50, initial api.md

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,0 +1,36 @@
+# ssh_scan API documentation
+
+**API Version:** 1.0
+**Status:** not ready
+
+This document explains the ssh_scan Web API, intended to be used in the Web API for ssh_scan.
+
+## Protocol Calls
+
+### scan
+
+**API Call:** `scan` <br>
+**API Method:** `POST`
+
+Scans a target, with an optional port (default 22). Equivalent to `ssh_scan -t host -p port`.
+Parameters:
+* `target` (required), which can be a hostname or an IP address
+
+POST Parameters:
+* `port` on which to scan (optional)
+
+Examples:
+* `/api/v1/scan?host=www.mozilla.org`
+* `/api/v1/scan?host=115.223.33.44`
+  * `port=9999` (POST data)
+
+### __version__
+
+**API Call:** `__version__` <br>
+**API Method:** `GET`
+
+Returns version details for ssh_scan being used and API.
+
+Example:
+* `/api/v1/__version__`
+  * Returns `{ "ssh_scan_version": "0.0.1", "api_version": "1" }`


docs/api.md
```diff
@@ -0,0 +1,34 @@
...
+**API Version:** 1
+
+This document explains the ssh_scan Web API, intended to be used in the Web API for ssh_scan.
+
+## Protocol Calls
+
+### scan
+
+**API Call:** `scan` <br>
+**API Method:** `POST`
```
claudijd: Would "GET" verbs also be allowed or would it always be a post?

docs/api.md
```diff
@@ -0,0 +1,34 @@
...
+Parameters:
+* `target` (required), which can be a hostname or an IP address
+
+POST Parameters:
+* `port` on which to scan (optional)
+
+Examples:
+* `/api/v1/scan?host=www.mozilla.org`
+* `/api/v1/scan?host=115.223.33.44`
+  * `port=9999` (POST data)
```
claudijd: What is the return expectation when doing the post?

What happens when things go well?

What happens when I put "foobar" as a host parameter? (does it error? what do errors look like?)

docs/api.md
```diff
@@ -0,0 +1,34 @@
...
+
+### __version__
```
claudijd: Assuming this is a GET method, can we clarify?

docs/api.md
```diff
@@ -0,0 +1,34 @@
+# ssh_scan API documentation
+
+**Last Updated:** 2016-09-19 <br>
+**Author:** agaurav77@yahoo.com <br>
```
claudijd: What's the practical value of author in this context?  We have it checked into git, so author should be self-evident with history and blame.

docs/api.md
```diff
@@ -0,0 +1,34 @@
...
+**API Version:** 1
```
claudijd: Can API version be 1.0?  I say this because we might want incremental improvements without major version bumps.

claudijd:
I tried out GitHub new review feature, so please forgive me if this isn't as easy to work with.  Let me know if you have any specific questions.

docs/api.md
agaurav77: Good point, this is redundant. Will remove.

docs/api.md
agaurav77: Sure, makes sense.

docs/api.md
agaurav77: Yeah, there would be GETs. Technically, GETs are for retrievals. Imagine if we start caching the results as well, then a GET could retrieve cached results and a POST could do a new scan.

docs/api.md
agaurav77: This should do a ssh_scan in the backend, so it would show the JSON that ssh_scan returns. For errors, we can check the JSON for the "error" field, and show the corresponding text.

docs/api.md
agaurav77: Yes, this should be GET. Sorry forgot to mention. I will fix this in the amended commit.

agaurav77:
Github's review is great! I'm glad Github finally has this :)

docs/api.md
```diff
@@ -0,0 +1,35 @@
+# ssh_scan API documentation
+
+**API Version:** 1.0
+
+This document explains the ssh_scan Web API, intended to be used in the Web API for ssh_scan.
```
claudijd: I'm wondering if at first we should have some label that states that this is a work in progress or a draft/design since the API doesn't exist yet

agaurav77: Good idea, I'll add that.

agaurav77:
@claudijd What more should I do here? Thanks!

agaurav77:
Closing since we're now tracking this in the wiki.

PR #212, opened 2016-09-20
report unknown when ubuntu version cannot be identified

### LINKED ISSUE #205: Make sure Ubuntu CPE reports correctly when the version isn't matched to an OS version
Example case was when we saw a banner like this.... "SSH-2.0-OpenSSH_6.6.1p1 Ubuntu-8"

I fixed the fingerprint for this in #204, but to reproduce this bug we just need to see something like "SSH-2.0-OpenSSH_6.6.1p1 Ubuntu-9" and it will improperly show up in the JSON output like this...

    "os_cpe": "o:canonical:ubuntu:#<SSHScan::OS::Ubuntu::Version:0x007f8c9e9741e0>",

agaurav77:
Should fix #205.

### DIFF
--- a/lib/ssh_scan/os/ubuntu.rb
+++ b/lib/ssh_scan/os/ubuntu.rb
@@ -200,7 +200,7 @@ module SSHScan
       end
 
       def cpe
-        "o:canonical:ubuntu:#{@version}"
+        "o:canonical:ubuntu" + (@version.to_s ? ":#{@version}" : "")
       end
     end
   end
--- a/spec/ssh_scan/banner/os/ubuntu_spec.rb
+++ b/spec/ssh_scan/banner/os/ubuntu_spec.rb
@@ -238,6 +238,10 @@ describe SSHScan::Banner do
         :ssh_lib_class => SSHScan::SSHLib::OpenSSH,
         :ssh_lib_version => "5.9p1",
       },
+      "SSH-2.0-OpenSSH_6.6.1p1 Ubuntu-9" => {
+        :os_class => SSHScan::OS::Ubuntu,
+        :os_cpe => "o:canonical:ubuntu",
+      }
     }
     checkFingerprints(expectations)
   end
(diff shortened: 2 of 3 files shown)


lib/ssh_scan/os/ubuntu.rb
```diff
@@ -3,7 +3,11 @@ module OS
     class Ubuntu
       class Version
         def initialize(version_string)
-          @version_string = version_string
+          if version_string.nil?
+            @version_string = "unknown"
```
claudijd: What if version_string was just nil rather than unknown?

I'm considering removing "unknown" references that have been a convention thus far because they don't follow CPE conventions.

spec/ssh_scan/banner/os/ubuntu_spec.rb
```diff
@@ -238,6 +238,11 @@
         :ssh_lib_class => SSHScan::SSHLib::OpenSSH,
         :ssh_lib_version => "5.9p1",
       },
+      "SSH-2.0-OpenSSH_6.6.1p1 Ubuntu-9" => {
+        :os_class => SSHScan::OS::Ubuntu,
+        :os_version => "unknown",
```
claudijd: Because a #to_s on a nil is "" that would be the switch here.

claudijd:
Looks good, just a comment about phasing out the "unknown" convention.

lib/ssh_scan/os/ubuntu.rb
agaurav77: How about making the cpe "o:canonical:ubuntu" (without ":unknown") then?

agaurav77:
@claudijd added another comment

lib/ssh_scan/os/ubuntu.rb
claudijd: yeah, that seems most logical.

agaurav77: Done!

Issue #211, opened 2016-09-20
make gem version checking an opt-out feature


agaurav77:
We should allow the user to opt out from gem version checking by using a flag like --supress-update-status.

claudijd:
This should also include exception handling for when access to rubygems is not available and we can't determine if an update is available.

claudijd:
See here for more context about the feature https://github.com/mozilla/ssh_scan/pull/210

agaurav77:
Closing, nothing more needed I guess.

PR #215, opened 2016-09-21
make update checking an opt-out feature

agaurav77:
For #211.

### DIFF
--- a/bin/ssh_scan
+++ b/bin/ssh_scan
@@ -113,6 +113,10 @@ opt_parser = OptionParser.new do |opts|
     options[:fingerprint_database] = fingerprint_db
   end
 
+  opts.on("--suppress-update-status", "Do not check for updates") do
+    options[:suppress_update_status] = true
+  end
+
   opts.on("-u", "--unit-test [FILE]",
           "Throw appropriate exit codes based on compliance status") do
     options[:unit_test] = true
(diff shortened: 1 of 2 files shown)


bin/ssh_scan
```diff
@@ -190,11 +194,19 @@ unless File.exists?(options[:policy])
 end
 
 # Check to see if we're running the latest released version
-update = SSHScan::Update.new
-if update.newer_gem_available?
-  options[:logger].warn("You're NOT using the latest version of ssh_scan, try 'gem update ssh_scan' to get the latest")
-else
-  options[:logger].info("You're using the latest version of ssh_scan #{SSHScan::VERSION}")
+if !options[:suppress_update_status]
```
claudijd: "if !" == "unless"

lib/ssh_scan/update.rb
```diff
@@ -5,6 +5,10 @@
 
 module SSHScan
   class Update
+    def initialize
+      @errors = []
+    end
```
claudijd: If we're adding this, we should probably make use of it.  Otherwise, maybe it can wait until we want to do use it.

bin/ssh_scan
agaurav77: Oh, right, thanks!

lib/ssh_scan/update.rb
agaurav77: I'm using this in bin/ssh_scan, so the user sees an error message whenever an exception occurs (say we cannot connect to rubygems or something failed).

claudijd: Oh, I missed that, sorry

Issue #216, opened 2016-09-21
Extend ssh fingerprints to include DSA


claudijd:
This means namespacing this to DSA and RSA....

    "fingerprints": {
      "dsa": {
        "md5": "9b:82:fe:2f:5d:bb:ac:13:86:10:3a:bb:4c:f2:ba:e5",
        "sha1": "f5:5c:78:f7:47:da:a0:bd:de:65:a2:bb:18:e3:ae:23:81:10:35:71",
        "sha256": "c8:e5:dd:19:0e:84:2b:ff:96:12:9a:e1:bd:66:c9:9e:80:02:69:5e:0d:58:b7:97:2e:d8:96:e5:c1:8d:c9:05"
      }
      "rsa": {
        "md5": "9b:82:fe:2f:5d:bb:ac:13:86:10:3a:bb:4c:f2:ba:e5",
        "sha1": "f5:5c:78:f7:47:da:a0:bd:de:65:a2:bb:18:e3:ae:23:81:10:35:71",
        "sha256": "c8:e5:dd:19:0e:84:2b:ff:96:12:9a:e1:bd:66:c9:9e:80:02:69:5e:0d:58:b7:97:2e:d8:96:e5:c1:8d:c9:05"
      }
    },

And extending the fingerprint DB to support this too, which might require a new row in the fingerprint DB to track key type as well as output modifications to keep the key types separate.

claudijd:
Nice job @royalharsh

royalharsh:
Thanks

PR #235, opened 2016-10-04
Extend ssh fingerprints to include DSA

royalharsh:
Extended fingerprints to include DSA. @claudijd ? https://github.com/mozilla/ssh_scan/issues/216


claudijd:
@royalharsh can you provide some testing example output to demonstrate both the RSA and DSA fingerprints on a single host?  I tried replicating and could replicate RSA behavior, but not DSA.

Example:

    "fingerprints": {
      "rsa": {
        "md5": "1a:8f:af:7d:4d:f7:2e:9d:e7:69:bb:07:17:4c:b0:71",
        "sha1": "b1:0c:5c:58:09:eb:20:3f:10:bc:e1:8b:0e:d8:fd:55:a8:21:60:d4",
        "sha256": "8a:92:84:06:ca:08:64:85:28:a0:36:0f:82:f3:dc:75:01:fc:01:82:35:e8:93:28:9d:e3:aa:b0:33:59:61:16"
      }
    },

.gitignore
```diff
@@ -29,7 +29,7 @@ build/
 
 # for a library or gem, you might want to ignore these files since the code is
 # intended to run in multiple environments; otherwise, check them in:
-# Gemfile.lock
+Gemfile.lock
```
claudijd: Thanks, long overdue.

lib/ssh_scan/crypto.rb
```diff
@@ -5,21 +5,37 @@ module Crypto
...
+          @data_string = OpenSSL::ASN1::Sequence([
+           OpenSSL::ASN1::Integer.new(@key.public_key.p),
+           OpenSSL::ASN1::Integer.new(@key.public_key.q),
+           OpenSSL::ASN1::Integer.new(@key.public_key.g)
+          ])
+        end
       end
 
-      def check_supported
+      def check_rsa_supported
```
claudijd: in ruby, if you're returning binary result, it's customary to make it a question.

How about "rsa_supported?" then you don't need to keep instance variables with the result.

claudijd: In a case of a high latency action, you can store the result in an instance variable on the class and allow the method to access it, but in this case we're just checking that the key exists and that it's of a specific class, so this should be reasonably performant.

lib/ssh_scan/scan_engine.rb
```diff
@@ -71,13 +71,26 @@ def scan_target(socket, opts)
...
+        end
+
+        if pkey.is_dsa_supported?
+          result['fingerprints'].merge!({
+            "dsa" => {
+              "md5" => pkey.fingerprint_md5,
+              "sha1" => pkey.fingerprint_sha1,
+              "sha256" => pkey.fingerprint_sha256,
+            }
+          })
```
claudijd: I'm beginning to wonder if we need to make multiple connections to determine RSA and DSA keys on the same server, so I asked for example that shows both.  I feel like that will tease out whether that's needed or not.

claudijd: Perhaps we'll need to make multiple net-ssh calls with options set to explicitly say we want this type of key and sort of brute the methods out, I'm really not sure yet.

royalharsh:
@claudijd @agaurav77 Here's the link to the output : [http://ix.io/1x0K](http://ix.io/1x0K) .

claudijd:
I pullet this down and tested on the same host in your example and got the following, which didn't include the fingerprints at all....

https://gist.github.com/claudijd/ac7a351892abc80b7fab3b6d8c5ae792

claudijd:
FWIW, I also tested on current master (32fde4da2f17e59874295d0387f4e67a657a304e) and didn't get the fingerprints label, so maybe we're missing some error reporting here.

claudijd:
I did some debugging here and from my view the host_keys are empty, which explains why we're seeing the fingerprints attribute and not the elements inside.  I do wonder if this has to do with net-ssh being able to establish a viable session or not, but it's silently erroring.

claudijd:
Yeah, this seems to be an issue with net-ssh on my end with retrieval of host-keys....

    $ ./bin/ssh_scan -t 52.74.215.8
    I, [2016-10-19T13:34:28.051140 #72572]  INFO -- : You're using the latest version of ssh_scan 0.0.15
    I, [2016-10-19T13:34:28.052079 #72572]  INFO -- : Started ssh_scan of 52.74.215.8

    From: /Users/jclaudius/code/ssh_scan/lib/ssh_scan/scan_engine.rb @ line 60 SSHScan::ScanEngine#scan_target:

        55:         raise SSHScan::Error::ClosedConnection.new if net_ssh_session.closed?
        56:         auth_session = Net::SSH::Authentication::Session.new(net_ssh_session, :auth_methods => ["none"])
        57:         auth_session.authenticate("none", "test", "test")
        58: 
        59:         require 'pry'
     => 60:         binding.pry
        61: 
        62:         result['auth_methods'] = auth_session.allowed_auth_methods
        63:         host_keys = net_ssh_session.host_keys
        64:         net_ssh_session.close
        65:       rescue Net::SSH::ConnectionTimeout => e

    [1] pry(#<SSHScan::ScanEngine>)> net_ssh_session.host_keys
    => []

claudijd:
@royalharsh what version of ruby and net-ssh are you running? I'm wondering if we need to set a minimum supported version to ensure this works for everyone.

    $ gem list | grep net-ssh
    net-ssh (3.2.0)

    $ ruby --version
    ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin14

royalharsh:
This is strange. I can still see fingerprints. 
```ruby
2.3.1 :003 > host_keys = net_ssh_session.host_keys
 => [#<OpenSSL::PKey::RSA:0x0000000289aab0>, #<OpenSSL::PKey::DSA:0x0000000289a3f8>] 
``` 
My tests were run using:
```bash
$ gem list | grep net-ssh
net-ssh (3.2.0)

$ ruby --version
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-linux]
```

royalharsh:
In your output, it does show the server host key algorithms: 
```
"server_host_key_algorithms": [
      "ssh-rsa",
      "ssh-dss"
    ],
```

royalharsh:
Does this have to do with the `~/. ssh/known_hosts ` file? I think I ssh in my server that's why I can only see the host_keys and not you.

claudijd:
@royalharsh I doubt that would be the issue, unless net-ssh is picking them up on the fly.  Can you try from a VM and see if you can replicate hostkeys or no hostkeys?

royalharsh:
I tried from another VM and got no hostkeys. 
```ruby
2.3.1 :003 > net_ssh_session.host_keys
 => []
```
My tests were run using:
```bash
$ gem list | grep net-ssh
net-ssh (3.2.0)

$ ruby --version
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-linux]
```

royalharsh:
I tried this: 
First, I scanned `192.30.253.113`, here's the result: [http://ix.io/1y9D](http://ix.io/1y9D) and I see the fingerprints.
```
"fingerprints": {
      "rsa": {
        "md5": "f6:ce:1e:d1:87:52:2d:26:81:89:ec:e6:d3:8f:9f:69",
        "sha1": "4d:b1:29:4f:e8:de:1c:14:1f:11:47:45:4e:8c:9a:c8:a8:45:76:40",
        "sha256": "9e:16:bf:1a:99:1b:b9:f6:9a:f9:24:fd:b2:55:29:51:d8:d2:20:06:14:c8:03:b0:bf:bb:95:9a:d1:a7:83:0f"
      }
    },
```
Then, I commented out `192.30.253.113` from my `~/.ssh/known_hosts` and here's the output: [http://ix.io/1y9E](http://ix.io/1y9E)
```
"fingerprints": {
    },
```
The second time it couldn't detect any fingerprints.

claudijd:
@royalharsh that's a pretty big tell that your example host is not letting you enumerate fingerprints.  We do know that we have the ability to enumerate fingerprints based on other testing examples.  Maybe it would be worth while to use shodan to find a couple more examples to test against.  If you can get them to return fingerprints (ideally both RSA/DSA) then I think we can land this.  Just provide a gist or pastebin link of the results so it's clear what you found.

royalharsh:
This should work now. Here's the output: [gist](https://gist.github.com/royalharsh/78fbd5d1f4d16dc323e4f139826f83e4)

claudijd:
Perfect, I was just able to confirm the behavior you showed in your gist.  Will land this shortly.

Issue #201, opened 2016-09-20
Make ssh_scan portable "binaries"


claudijd:
A common problem with ruby apps is not having a native ruby environment where you want to run it.  I looked into options for this and traveling-ruby seems like the right fit.  It basically allows you to ship the entire application and a pre-compiled ruby with the file.

There is a tutorial here, which I think we'd want to test out a bit and then just turn into a rake task that we run on master or just at release time.

https://github.com/phusion/traveling-ruby/blob/master/TUTORIAL-1.md

I went through the hello world tutorial and it seemed to be pretty promising.  Would be cool if someone was willing to prove that all the way out and send a PR for it.

yashmehrotra:
@claudijd Hi, I would like to work on this one, can you assign it to me ?

jinankjain:
@yashmehrotra I was trying to work but got busy with some other work. There is kind of one bug with my rakefile if you could fix that up it is good to go.

```
require 'rubygems'
require 'rake'
require 'rubygems/package_task'
require 'rspec'
require 'rspec/core'
require 'rspec/core/rake_task'
require 'bundler/setup'

$:.unshift File.join(File.dirname(__FILE__), "lib")

require 'ssh_scan'

task :default => :spec

desc "Run all specs in spec directory"
RSpec::Core::RakeTask.new(:spec)

PACKAGE_NAME = "ssh_scan"
VERSION = "0.0.15"
TRAVELING_RUBY_VERSION = "20150210-2.1.5"
SQLITE3_VERSION = "1.3.9"  # Must match Gemfile

desc "Package your app"
task :package => ['package:linux:x86', 'package:linux:x86_64', 'package:osx']

namespace :package do
  namespace :linux do
    desc "Package your app for Linux x86"
    task :x86 => [:bundle_install,
      "packaging/traveling-ruby-#{TRAVELING_RUBY_VERSION}-linux-x86.tar.gz",
      "packaging/traveling-ruby-#{TRAVELING_RUBY_VERSION}-linux-x86-sqlite3-#{SQLITE3_VERSION}.tar.gz"
    ] do
      create_package("linux-x86")
    end

    desc "Package your app for Linux x86_64"
    task :x86_64 => [:bundle_install,
      "packaging/traveling-ruby-#{TRAVELING_RUBY_VERSION}-linux-x86_64.tar.gz",
[...]

PR #237, opened 2016-10-18
Rake tasks for packaging portable binaries

rishabhs95:
Rake tasks for generating portable `ssh_scan` binaries. Supports binaries for linux x86, linux x86_64, osx and windows. Uses travelling-ruby for packaging ruby with `ssh_scan`. 
Issue #201


claudijd:
@rishabhs95 nice job on this, I made one small tweak to get the version information from ssh_scan/version rather than it being hard coded.  We can talk more later about how this should fit into our release process.

rishabhs95:
@claudijd yeah sure, getting version from ssh_scan/version would be better

claudijd:
https://github.com/mozilla/ssh_scan/blob/master/Rakefile#L20 :)

Issue #236, opened 2016-10-05
Add something to enforce commit signing on ssh_scan


claudijd:
Idea Reference: https://github.com/escapewindow/cot-gpg-keys/blob/master/check_commit_signatures.py

agaurav77:
Even if we have code that can check whether the latest commit is signed or not, where do we add it? 

Do we add a test, say which ensures that each time a travis build is done, the commit must be signed? If we go this way, this would mean that any contributor who does not have signed the last commit cannot do a PR, since the travis build would fail.

Good idea though. I am willing to do this in Ruby if I know where to add the check.

claudijd:
We could just restrict the checks to master, so if there is an unsigned head on master than the master will begin to fail after merge.  This would alert us that something bad has happened.

royalharsh:
How about using bot accounts ? Bots would check if commits have been signed or not and then add labels accordingly, this would allow us to keep a check if commits have been signed or not before merging. One example of such bot : [clabot](https://clabot.github.io/)

claudijd:
@royalharsh this seems cool, I was hoping to have something that could be integrated into the PR logic so it makes it clear whether all commits in the chain are signed.  GitHub already makes this clear with the verified labels, but I'd like something to enforce it for anything that lands in master.  I suppose this could make it harder for landing code by mergers, but I'd like to at least explore the options so we can show others it's possible and use ssh_scan as a reference architecture for it.

claudijd:
#242 is now in production, and seems to have worked

https://travis-ci.org/mozilla/ssh_scan/jobs/169247977

I'll be sending a PR test shortly that is not signed and we'll see how it works.

claudijd:
unsigned commit test here => https://github.com/mozilla/ssh_scan/pull/244

agaurav77:
I think I should change
```
git log --no-merges --format='%H:%GG'
```
to
```
git log --no-merges --format='%H:%GG' master
```

The current enforcing logic always checks the current branch, and we want to fail when master seems fishy.

claudijd:
it did not seem to fail here: https://travis-ci.org/mozilla/ssh_scan/jobs/169346670

agaurav77:
I fixed a few things again. 

Since the last commit is unsigned (just to test), the next commit should fail on Travis.

agaurav77:
The next commit will be see the latest master as signed. So, Travis should not fail.

agaurav77:
[...]

PR #242, opened 2016-10-20
enforce signed commits on master

agaurav77:
For #236 

Still have to add pwnbus and kang's GPG keys.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,12 @@ matrix:
   include:
   - rvm: ruby-head
     env:
+      - LABEL=enforce_signed_commits
+    script:
+      - chmod 755 ./spec/ssh_scan/enforce_signed_commits.rb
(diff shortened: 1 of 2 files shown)


claudijd:
@agaurav77 this looks interesting, I'd like to land it now and have pwnbus or kang be our control set to test the effectiveness of the control.  If it doesn't work the way we want it to, we'll iterate on it.  I also have some curiousity on how external contribs will react to this or whether a signed merge will make it ok.

claudijd:
Additionally, since the tool we're using to verify also has it's source in the repo, it's very possible that a malicious person could simply overwrite this script to get it to pass.  Either way, I think this is a good tripwire to have and will help us get better about signing all of our commits, so I think it's still in a win in my book.

agaurav77:
@claudijd Sounds good, please give it a go.

PR #244, opened 2016-10-20
Unsigned test commit

claudijd:
This is a verification step to ensure the signed commits test is working as expected.


claudijd:
ok, merged, this should cause a failure on master when it builds, since it's head has been update.

Issue #49, opened 2016-08-03
Create a queueing/job/workload engine for ssh_scan


claudijd:
Currently, this tool is run a single IP at a time.

I'd like to add some sort of queueing/workload component that would allow ssh_scan to be offered up as a service that could serve multiple clients via some sort of API functionality.

This would open up the users of the system to select the client-side language of their choice to manage their own scans.

claudijd:
The above would need to be highly scalable such that it could scan to thousands of ssh instances

claudijd:
Some seed ideas here include:

- Redis, RabbitMQ, etc.

PR #259, opened 2016-11-02
job_queue - worker api model

rishabhs95:
I was also working on the job queueing model and trying to implement it. The implementation is not complete at the moment. I will debug and improve the implementation tomorrow, if It looks like the right direction. @agaurav77 @royalharsh 
Issue #49

### DIFF
--- a/lib/ssh_scan/worker.rb
+++ b/lib/ssh_scan/worker.rb
@@ -2,10 +2,8 @@ require 'ssh_scan/scan_engine'
 
 module SSHScan
   class Worker
-    scan_engine = SSHScan::ScanEngine.new()
-
     def process_job(job)
-      scan_engine.scan(job).to_json
+      SSHScan::ScanEngine.scan(job).to_json
     end
   end
 end
--- a/lib/ssh_scan/worker.rb
+++ b/lib/ssh_scan/worker.rb
@@ -2,8 +2,15 @@ require 'ssh_scan/scan_engine'
 
 module SSHScan
   class Worker
-    def process_job(job)
-      SSHScan::ScanEngine.scan(job).to_json
+    @scan_engine = SSHScan::ScanEngine.new
+    @logger = Logger.new(STDOUT)
+
+    def self.process_job(job)
+      @logger.info("SSHScan::ScanEngine started for job: #{job[:uuid]}")
+      scan_result = scan_engine.scan(job).to_json
+      puts scan_result.inspect
+      @logger.info("SSHScan::ScanEngine finished for job: #{job[:uuid]}")
+      return scan_result
     end
   end
 end
--- a/lib/ssh_scan/worker.rb
+++ b/lib/ssh_scan/worker.rb
@@ -1,0 +1,11 @@
+require 'ssh_scan/scan_engine'
+
+module SSHScan
+  class Worker
+    scan_engine = SSHScan::ScanEngine.new()
+
+    def process_job(job)
+      scan_engine.scan(job).to_json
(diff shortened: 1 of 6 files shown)


lib/ssh_scan/worker.rb
```diff
@@ -0,0 +1,44 @@
+require 'thread'
+
+module SSHScan
+  class Worker
```
claudijd: Consider maybe renaming this to imply multiple workers.  WorkerPool?

claudijd: Queue?

claudijd:
As I'm thinking about this more, I think perhaps we're merging a couple functions that should district entities, I'm going to draft a PR as a suggestion so we can contrast them and see what you think fits better.

claudijd:
Suggestion PR to better show how I think the interfaces could a bit cleaner.

https://github.com/mozilla/ssh_scan/pull/261

It shows how we could make these internal to start with (so we have something functional) and then it makes it pretty easy to add the endpoints to the API to push the workers out into their own independent process that just use the API.

rishabhs95:
I've made some changes with respect to #261, I'm further looking into improvements as well but just go through and drop some comments on the code @claudijd @agaurav77 @royalharsh @Rajat-Goyal

claudijd:
I'm biased here, but I do feel like this gets us in a direction that allows us to incrementally add and improve the feature set gradually.

claudijd:
Also, the automated CI jobs are failing on this PR.  I believe the reason is that there are two issues...

1.) api.rb is failing due to a end syntax error (this can be fixed by adding and "end" at the end of the file)
2.) there are unit-tests failing on API expectations but the usage doesn't match the API documentation (example: they are using target and port in POST body, but target is a URI param according to the api docs)

rishabhs95:
I've updated the PR with respect to the new wiki docs, i was trying debugging this PR, logging the statements show that the error is in calling the Worker's call to `scan_engine` as our control doesn't reaches the `scan_engine.scan()`.

I am busy with my exams for the moment, will get back to it as soon as I get some time 🤓

claudijd:
I tried doing some testing of this branch and got the following...

1. Made a request to task a scan using Curl**
$ curl -k -X POST https://127.0.0.1:8000/api/v0.0.1/scan?target=ssh.mozilla.com
2. Got a scan uuid back from the host in JSON format
{"uuid":"2b848a55-8420-4bf6-8fc0-f94c1aaae8c2"}
3. Poll the API for the status of that particular scan ID using Curl
 curl -ki https://127.0.0.1:8000/api/v0.0.1/scan/results?uuid=2b848a55-8420-4bf6-8fc0-f94c1aaae8c2
4. Got an empty 200 response with no actual content returned

**Note:** I also observed that via the application logs that an empty array was being printed in the logs...

**Example log entry:**  
    W, [2016-11-30T14:50:04.328462 #23939]  WARN -- : JobQueue Size: 0
    []
    W, [2016-11-30T14:50:06.333474 #23939]  WARN -- : JobQueue Size: 0

rishabhs95:
Yeah, I put some debug statements and the problem seems like the `Worker` class's call to scan_engine as our control doesn't reaches the `scan_engine.scan()`. Just looking around for the errors

claudijd:
@rishabhs95 I'm looking at it in more detail now, give me a bit I'll get you unstuck

rishabhs95:
cool, that will help @claudijd 👍 
sorry, I couldn't give it much time this week due to the exams

Issue #233, opened 2016-10-04
Follow up API improvements


claudijd:
- Add sane security headers to API (to make httpobs happy)
  - https://wiki.mozilla.org/Security/Guidelines/Web_Security
  - Secureheaders is a good starting point (https://github.com/twitter/secureheaders)
- Sync up API with API docs (host vs. target issue)
- Add custom handler for 404s (Ref: http://stackoverflow.com/questions/19781051/rendering-404-in-sinatra-if-file-not-found) and make 404 reference API documentation
- Support self-signed certs (it's not ideal, but it's better and starts to lay framework for SSL/TLS expectations by default)
- Add custom server header
- Add contribute.json
- Add Robots.txt

claudijd:
#234 addresses custom 404 handler and starts using Twitter's secure_headers gem to give us full control over all the security headers we'd want to set.  I just did in enough there to get it functional, but the secure header config should be reworked.

claudijd:
custom server header and moar sane security header settings added to #234

agaurav77:
Changed the wiki, so that's done.

claudijd:
@agaurav77 no SSL at all at this point, would be nice to support at least self-signed ssl vs http as the API default and then provide a path for letsencrypt support.

claudijd:
This has been resolved.

PR #260, opened 2016-11-03
add https support 

Rajat-Goyal:
Add support for https, #233  
I and @agaurav77 are still working on more suitable tests.

### DIFF
--- a/bin/ssh_scan_api
+++ b/bin/ssh_scan_api
@@ -22,6 +22,18 @@ opt_parser = OptionParser.new do |opts|
     exit
   end
 
+  opts.on("--https", "Use HTTPS") do
+    options[:https] = true
+  end
+
+  opts.on("--https-key KEY", "Specify the path to your KEY file for your server") do |key|
+    options[:key] = File.realpath(key)
+  end
+
+  opts.on("--https-crt CRT", "Specify the path to your CRT file for your server") do |crt|
+    options[:crt] = File.realpath(crt)
+  end
+
   opts.on_tail("-h", "--help", "Show help") do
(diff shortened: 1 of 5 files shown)


bin/ssh_scan_api
```diff
@@ -22,6 +22,18 @@ opt_parser = OptionParser.new do |opts|
     exit
   end
 
+  opts.on("--https", "Use HTTPS") do
+    options[:https] = true
+  end
```
claudijd: Does this just serve a self-signed cert if the key and crt files aren't provided?

claudijd: I tried running the following and it didn't serve a self-signed cert by default....

./bin/ssh_scan_api --https

claudijd:
I'm making some slight changes to this, but will be landing shortly

bin/ssh_scan_api
agaurav77: https would work like this
./bin/ssh_scan_api --https --https-key <keyfile> --https-crt <crtfile>

We were initially going to go for generation of self signed certs using OpenSSL everytime, but we thought you might want to add in one of LetsEncrypt based certificates on your own. If you want certs to be generated everytime we can add that back.

claudijd:
Changes I made can be found here, basically don't provide an HTTP option at all.

By default the API will be served as self-signed as generated by the library and you can override the cert/key using the options you added to the api command line.  This should make it easy for people to use SSL in the future.

https://github.com/mozilla/ssh_scan/commit/bbbbb6854510f138d65bd0535cab7585186e1c9b

PR #261, opened 2016-11-03
Prototype example of queue/worker abstraction shims

claudijd:
How to load the JobQueue and watch it expand and contract...

curl -k -X POST https://127.0.0.1:8000/api/v0.0.1/scan?target=www.mozilla.org

Run it many times to watch the queue load up.

### DIFF
--- a/lib/ssh_scan/api.rb
+++ b/lib/ssh_scan/api.rb
@@ -3,6 +3,7 @@ require 'sinatra/namespace'
 require 'ssh_scan/version'
 require 'ssh_scan/policy'
 require 'ssh_scan/scan_engine'
+require 'ssh_scan/job_queue'
 require 'json'
 require 'haml'
 require 'secure_headers'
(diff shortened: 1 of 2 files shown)


claudijd:
@royalharsh @rishabhs95 can you guys have a look at this and let me know what you think?

lib/ssh_scan/api.rb
```diff
@@ -99,25 +123,15 @@ def add(*args, &block)
...
-          :timeout => 2,
-          :verbosity => nil,
-          :logger => NullLogger.new,
-          :fingerprint_database => "fingerprints.db",
-        }
-        options[:sockets] << "#{params[:target]}:#{params[:port] ? params[:port] : "22"}"
-        options[:policy_file] = SSHScan::Policy.from_file(options[:policy])
-        scan_engine = SSHScan::ScanEngine.new()
-        scan_engine.scan(options).to_json
+        # TODO: there needs to be some sort of ID tracking here so the user can come back later and retrieve the results of their scan
```
royalharsh: I think we'd need some sort of database if we are going to let users retrieve the results  of their previous scans.

royalharsh:
How to decide the number of threads we can use while scanning multiple ips ?

claudijd:
Close this, as it looks like the code was ported and adapted into #259

PR #263, opened 2016-12-01
Rishabhs95 job queueing suggestions

claudijd:
An addendum to #259 to help suggest the direction of abstracting the worker class.  Debugging threads within threads was just a pain to deal with.

/cc @rishabhs95 please check these changes out. I cut out some code (namely the fingerprint db) and abstracted the worker class into it's own process and added the API endpoints necessary for the worker to remotely interact with the API.  It's still very much a work in progress, but I though this would help get you unstuck on #259

### DIFF
--- a/bin/ssh_scan
+++ b/bin/ssh_scan
@@ -11,14 +11,14 @@ require 'logger'
 
 #Default options
 options = {
-  :sockets => [],
-  :policy => File.expand_path("../../policies/mozilla_modern.yml", __FILE__),
-  :unit_test => false,
-  :timeout => 2,
-  :threads => 5,
-  :verbosity => nil,
-  :logger => Logger.new(STDERR),
-  :fingerprint_database => "./fingerprints.db"
+  "sockets" => [],
+  "policy" => File.expand_path("../../policies/mozilla_modern.yml", __FILE__),
+  "unit_test" => false,
+  "timeout" => 2,
+  "threads" => 5,
+  "verbosity" => nil,
+  "logger" => Logger.new(STDERR),
+  "fingerprint_database" => "./fingerprints.db"
 }
 
 target_parser = SSHScan::TargetParser.new()
@@ -31,7 +31,7 @@ opt_parser = OptionParser.new do |opts|
           "IP/Ranges/Hostname to scan") do |sockets|
     sockets.each do |socket|
       ip, port = socket.chomp.split(':')
-      options[:sockets] += target_parser.enumerateIPRange(ip, port)
+      options["sockets"] += target_parser.enumerateIPRange(ip, port)
     end
   end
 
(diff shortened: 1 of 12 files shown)


bin/ssh_scan_worker
```diff
@@ -0,0 +1,20 @@
...
+$:.unshift File.join(File.dirname(__FILE__), "../lib")
+
+require 'ssh_scan'
+
+# An example of a dummy worker process performing scans
+logger = Logger.new(STDOUT)
+
+loop do
+  sleep 0.5
+  worker = SSHScan::Worker.new({:server => "192.168.10.37", :port => 8000})
```
pwnbus: Does it make sense to throw the "server" into a config file instead of hardcoded here? I'm assuming this was a leftover artifact from testing?

claudijd: yeah, prototype code, eventually it should just read from a config.  This PR is really just to help get over the hump of troubleshooting, which was getting unweildy without having separate a separate process for the worker logic.

bin/ssh_scan_worker
```diff
@@ -0,0 +1,20 @@
...
+  response = worker.get_work
+  if response["work"]
+    job = response["work"]
+    results = worker.perform_work(job)
+    worker.post_results(results, job)
```
claudijd: This logic should all pretty much live within the worker class.

We shouldn't need to expose all this sausage making.

bin/ssh_scan_worker
pwnbus: Understood.

bin/ssh_scan
```diff
@@ -11,14 +11,14 @@ require 'logger'
...
-  :logger => Logger.new(STDERR),
-  :fingerprint_database => "./fingerprints.db"
+  "sockets" => [],
+  "policy" => File.expand_path("../../policies/mozilla_modern.yml", __FILE__),
+  "unit_test" => false,
+  "timeout" => 2,
+  "threads" => 5,
+  "verbosity" => nil,
+  "logger" => Logger.new(STDERR),
+  "fingerprint_database" => "./fingerprints.db"
```
claudijd: Switched these from symbols to string refs to make it easier from the JSON <=> Hash changes introduced with Worker API abstraction.  Let data mangling is better.

lib/ssh_scan/scan_engine.rb
```diff
@@ -133,34 +135,34 @@ def scan(opts)
...
+      #       host_key_algo.values.each do |fingerprint|
+      #         fingerprint_db.find_fingerprints(fingerprint).each do |other_ip|
+      #           next if ip == other_ip
+      #           result['duplicate_host_key_ips'] << other_ip
+      #         end
+      #       end
+      #     end
+      #     result['duplicate_host_key_ips'].uniq!
+      #   end
+      # end
```
claudijd: Removed fingerprinting from here temporarily, we'll need to further abstract fingerprint and compliance decorators so they are decorated by the API itself and not workers.

lib/ssh_scan/worker.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      uri = URI("https://#{@server}:#{@port}/api/v0.0.1/work?worker_id=#{@worker_id}")
+      http = Net::HTTP.new(uri.host, uri.port)
+      http.use_ssl = true
+      http.verify_mode = OpenSSL::SSL::VERIFY_NONE # unless @verify_ssl == true
+      options_mask = OpenSSL::SSL::OP_NO_SSLv2 + OpenSSL::SSL::OP_NO_SSLv3 + OpenSSL::SSL::OP_NO_COMPRESSION
+      http.ssl_options = options_mask
+      request = Net::HTTP::Get.new(uri.path)
+      response = http.request(request)
+      JSON.parse(response.body)
```
claudijd: This is missing a lot of exception handling and will just die if it runs into a bad HTTP response.

spec/ssh_scan/api_spec.rb
```diff
@@ -27,20 +27,20 @@ def app
...
+  #
+  #   parsed_response_body = JSON.parse(last_response.body)
+  #
+  #   expect(parsed_response_body).to be_kind_of(::Array)
+  #   expect(parsed_response_body.first["ssh_scan_version"]).to eql(SSHScan::VERSION)
+  #   expect(parsed_response_body.first["ip"]).to eql(bad_ip)
+  #   expect(parsed_response_body.first["port"]).to eql(port)
+  #   expect(parsed_response_body.first["error"]).to match(/ConnectTimeout: (Connection|Operation) timed out - user specified timeout/)
+  #   expect(parsed_response_body.first["hostname"]).to eql("")
+  # end
```
claudijd: I commented this because I think the spec expectations are incorrect.

PR #270, opened 2016-12-01
Add authentication to ssh_scan_api

### LINKED ISSUE #262: Need a mechanism for API authentication
Currently, the API prototype has not authentication in place.  As an easy get started way of having authentication in our lives, I'd like to propose a config based token model.

The way it could work is we have a YAML file that sits on disk and it read by the ssh_scan_api.

We would have an authentication section and for each user we would have their api token.

Example:

authentication:
  harsh:
    token: 3206004d-15ea-4bb7-8e35-78e2d4594a54
  claudijd:
    token: af6c8a78-7b84-490a-a540-3f42ea905929

In addition to the configuration file defining the authentication tokens, we would also need to adapt the API routes to enforce those tokens and we would need to update the documentation to reflect which routes require auth and which ones don't.

claudijd:
Adds a crude authentication layer to the ssh_scan_api

Fixes #262

### DIFF
--- a/spec/ssh_scan/version_spec.rb
+++ b/spec/ssh_scan/version_spec.rb
@@ -22,14 +22,11 @@ describe SSHScan::VERSION do
   end
 
   it "SSHScan::API_VERSION should have 3 levels" do
-    expect(SSHScan::API_VERSION.split('.').size).to eql(3)
+    expect(SSHScan::API_VERSION.split('.').size).to eql(1)
   end
 
   it "SSHScan::API_VERSION should have a number between 1-20 for each octet" do
-    SSHScan::API_VERSION.split('.').each do |octet|
-      expect(octet.to_i).to be >= 0
-      expect(octet.to_i).to be <= 20
-    end
+    expect(SSHScan::API_VERSION).to eql("1")
   end
 
 end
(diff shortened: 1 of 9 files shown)


spec/ssh_scan/version_spec.rb
```diff
@@ -22,14 +22,11 @@
...
-    expect(SSHScan::API_VERSION.split('.').size).to eql(3)
+    expect(SSHScan::API_VERSION.split('.').size).to eql(1)
   end
 
   it "SSHScan::API_VERSION should have a number between 1-20 for each octet" do
-    SSHScan::API_VERSION.split('.').each do |octet|
-      expect(octet.to_i).to be >= 0
-      expect(octet.to_i).to be <= 20
-    end
+    expect(SSHScan::API_VERSION).to eql("1")
```
claudijd: API version was too long 👍

agaurav77:
A late question but still: how to set HTTP_SSH_SCAN_AUTH_TOKEN on the user side? If the user is doing a curl, can we specify a way to send tokens in the request somehow? Or does it work by setting an environment variable on the user's end (or a config file value)?

Issue #266, opened 2016-12-01
We need a simple data store for the API results


claudijd:
This was discussed in todays MWoS meeting, we're thinking SQLite to start, but would like the interfaces to be non-sqlite specific so we can swap in/out DB technology as demand requires.

@royalharsh @rishabhs95 are planning to tackle this in the next week

claudijd:
This has long been solved for, just a dangling issue.

PR #284, opened 2016-12-06
Added data store for API results

royalharsh:
For https://github.com/mozilla/ssh_scan/issues/266 . 

`/api/v1/scan/results?uuid=uuid_string` returns the scan with the specified uuid.
`/api/v1/work/results/worker_id/uuid` returns the scan with specified worker_id and uuid.
`/api/v1/scan/results/delete/worker_id/uuid` deletes the scan with specified worked_id and uuid.
`/api/v1/scan/results/delete/all` deletes all the rows in the table.

@claudijd @pwnbus @gdestuynder review?

### DIFF
--- a/lib/ssh_scan/api.rb
+++ b/lib/ssh_scan/api.rb
@@ -4,6 +4,7 @@ require 'ssh_scan/version'
 require 'ssh_scan/policy'
 require 'ssh_scan/job_queue'
 require 'ssh_scan/worker'
+require 'ssh_scan/api_db'
 require 'json'
 require 'haml'
 require 'secure_headers'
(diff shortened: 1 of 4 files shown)


lib/ssh_scan/api_db.rb
```diff
@@ -0,0 +1,61 @@
...
+        self.create_schema
+      end
+    end
+
+    def create_schema
+      @db.execute <<-SQL
+        create table api_schema (
+          worker_id varchar(100),
+          uuid varchar(100),
+          result json
```
agaurav77: Not too sure if json is a datatype with SQLite3 that Ruby has (It is a datatype in SQLite 3.9). Did you want to do varchar or longtext here?

lib/ssh_scan/api_db.rb
```diff
@@ -0,0 +1,61 @@
+require 'sqlite3'
+
+module SSHScan
+  class APIDatabaseHelper
```
claudijd: What if we just called it SSHScan::Database as the abstraction layer and then provide an SSHScan::Database::SQLite class?  The reason being is that I think SSHScan::Database should not know anything about SQLite so that we can create SSHScan::Database::Postgres, SSHScan::Database:MySQL, SSHScan::Database:Mongo, etc. without changing interfaces.

lib/ssh_scan/api.rb
```diff
@@ -97,8 +98,28 @@ class API < Sinatra::Base
       end
 
       get '/scan/results' do
-        # TODO: get a given scan result, by UUID and return it as JSON
-        '{"I am not finished yet"}'
+        uuid = params[:uuid]
+
+        if uuid.empty?
+          '{"I am not finished yet"}'
```
claudijd: I'm fixing this before landing

lib/ssh_scan/api_db.rb
claudijd: We may need to take a second pass on the schema for the DB, but as is, it works.  I would like to be able to query elements within the scan results eventually, but I think it could be a future improvement and not affect user/worker interfaces.

lib/ssh_scan/api_db.rb
claudijd: I'll probably propose that we do this in the future PR, just wanted to express the idea here.

Issue #287, opened 2016-12-09
Provide support for alternative Database types


claudijd:
See abstraction database comments from https://github.com/mozilla/ssh_scan/pull/284

claudijd:
Specific guidance that I would like to carry over here includes...

A single object that is used by SSHScan::API for data storage abstraction called SSHScan::Database

Based on options supplied to SSHScan::Database, the user will select their DB type.  DB types will be defined in the following name space like SSHScan::Database::SQLite, SSHScan::Database::MongoDB, SSHScan::Database::Postgres, etc.

The expectation is not that we implement each and every DB type, but that we design for the concept of multiple DB types and maybe implement at least one that better supports JSON so our API protype can easily generate stats.

claudijd:
This is done.

PR #291, opened 2016-12-13
Add Mongo as alternate database

royalharsh:
https://github.com/mozilla/ssh_scan/issues/287 
Now, we have to use `save=true` to save the results in the database and we can select database type by specifying `db=sqlite`, otherwise, it defaults to `db=mongodb`.

For example,
```bash
curl -k -X POST https://127.0.0.1:8000/api/v1/scan -d "target=ssh.mozilla.com&db=sqlite&save=true"
```


lib/ssh_scan/sqlite_db.rb
```diff
@@ -0,0 +1,64 @@
+require 'sqlite3'
+
+module   SSHScan
+  module Database
+    class SQLiteDb
```
agaurav77: rename to SQLite, Db is redundant (MongoDb is fine, its called like that)

agaurav77: Ideally, put sqlite_db.rb and mongo_db.rb in a directory called "database", then in database.rb, include both these files.

lib/ssh_scan/api_db.rb
```diff
@@ -1,61 +1,39 @@
-require 'sqlite3'
+require 'mongo'
```
agaurav77: rename to mongo_db.rb ?

agaurav77:
@royalharsh Tests? Would be nice to have the test functionality ready from the beginning.

royalharsh:
@agaurav77 Yeah will be adding tests soon.

lib/ssh_scan/database/mongo_db.rb
```diff
@@ -0,0 +1,39 @@
+require 'mongo'
+
+Mongo::Logger.logger.level = ::Logger::FATAL
+
+module   SSHScan
+  module Database
+    class MongoDb
```
claudijd: These abstractions turned out very well!

lib/ssh_scan/worker.rb
```diff
@@ -27,7 +26,13 @@ def run!
...
             job = response["work"]
             results = perform_work(job)
-            @api_db.add_scan(@worker_id, job["uuid"], results.to_json)
+            if job["db"].eql?("sqlite") && job["save"].eql?("true")
+              api_db = SSHScan::Database::SQLite.new
+              api_db.add_scan(@worker_id, job["uuid"], results)
+            elsif job["save"].eql?("true")
+              api_db = SSHScan::Database::MongoDb.new
+              api_db.add_scan(@worker_id, job["uuid"], results)
+            end
```
claudijd: I feel like the worker shouldn't have any need to know about the DB type.  Isn't that for the API to worry about?

lib/ssh_scan/api.rb
```diff
@@ -86,11 +86,22 @@ class API < Sinatra::Base
...
           :fingerprint_database => "fingerprints.db",
+          :db => "mongodb",
+          :save => "false"
         }
         options[:sockets] <<
           "#{params[:target]}:#{params[:port] ? params[:port] : "22"}"
         options[:policy_file] = options[:policy]
         options[:uuid] = SecureRandom.uuid
+        options[:db] = params[:db].downcase unless params[:db].nil?
+        options[:save] = params[:save] unless params[:save].nil?
```
claudijd: What's the vision for the "save" parameter? Shouldn't we save all results?

Also, it doesn't make sense to me why the client would specify why type of database the API should use.  Shouldn't DB type just that be transparent to the end user and it be a function of the API's configuration file?

claudijd:
I must say, I'm very excited to see the work that's been done here.  Great job on this!  I did have a couple areas where I think we can tighten things up.  Summarizing below:

1.) Workers don't need to know the DB type (it shouldn't be something they need to know), this is the responsibility of the API, if I understand it correctly.
2.) Clients don't need to know the DB type (it shouldn't be something they need to know), this is the responsibility of the API, if I understand it correctly.
3.) What's the vision for the "save" parameter, shouldn't we save all results?
4.) API DB type should ideally be driven by the API config (this relates directly to 1 and 2 above where I think we're getting a little fuzzy about interfaces and responsibilities)
5.) Also, the unit-tests mentioned above would be handy to make sure it doesn't break as we iterate on it.

lib/ssh_scan/database/sqlite_db.rb
```diff
@@ -0,0 +1,65 @@
...
+        @db.execute "insert into api_schema values ( ? , ? , ? , ? )",
+                    [worker_id, uuid, result.to_json, Time.now.to_s]
+      end
+
+      def delete_scan(worker_id, uuid)
+        @db.execute(
+          "delete from api_schema where worker_id = ? and uuid = ?",
+          worker_id, uuid
+        )
+      end
```
claudijd: Should you need to know the worker_id to delete a scan?  Functionally, I would think the uuid would be enough.

lib/ssh_scan/database/sqlite_db.rb
```diff
@@ -0,0 +1,65 @@
...
+          @db = ::SQLite3::Database.new(database_name)
+          create_schema
+        end
+      end
+
+      def create_schema
+        @db.execute <<-SQL
+          create table api_schema (
+            worker_id varchar(100),
+            uuid varchar(100),
```
claudijd: I'm wondering if in the schema we should have uuid, result, and then worker_id.  I say that because functionally the only reason we care about who did the work is if the work is somehow corrupted or problematic in some way, so I'm thinking of it more like an attribute rather than say a primary key item.

lib/ssh_scan/database/mongo_db.rb
```diff
@@ -0,0 +1,39 @@
...
+      end
+
+      def add_scan(worker_id, uuid, result)
+        @scans.insert_one("worker_id" => worker_id, "uuid" => uuid,
+                          "scan" => result.first, "scanned_on" => Time.now)
+      end
+
+      def delete_scan(worker_id, uuid)
+        @scans.delete_one(:worker_id => worker_id, :uuid => uuid)
+      end
```
claudijd: Same questions here, as I asked in SQLite abstraction about whether worker ID is relevant in this context.

lib/ssh_scan/database/mongo_db.rb
```diff
@@ -0,0 +1,39 @@
...
+  module Database
+    class MongoDb
+      def initialize
+        @db = Mongo::Client.new(['127.0.0.1:27017'], :database => 'api_db')
+        @scans = @db[:scans]
+      end
+
+      def add_scan(worker_id, uuid, result)
+        @scans.insert_one("worker_id" => worker_id, "uuid" => uuid,
+                          "scan" => result.first, "scanned_on" => Time.now)
```
claudijd: Same questions here, as I asked in SQLite abstraction about element ordering.

lib/ssh_scan/api.rb
```diff
@@ -140,7 +147,7 @@ class API < Sinatra::Base
         uuid = params['uuid']
 
         if worker_id.empty? || uuid.empty?
-          return {"accepted" => "false"}.to_json
+          return {"completed" => "false"}.to_json
         end
 
         settings.db.find_work_result(worker_id, uuid)
```
claudijd: Is this even necessary? I cannot think of why the API would need to retrieve a scan result based on worker_id and uuid.  Shouldn't uuid be enough?

claudijd:
Suggestion for format of DB configuration

https://gist.github.com/claudijd/c9c93fcd0107f9cabffc9e93eee5fab1

royalharsh:
1) Added a DB config file. 
2) `/scan/results/delete/:worker_id/:uuid` => `/scan/results/delete/:uuid`
3) Moved `__version__` out of `/api/v1` namespace. (Thanks to @rishabhs95) Previously, to know the API version, you would actually had to know version of API to make the call.
@claudijd @agaurav77 review?

lib/ssh_scan/api.rb
```diff
@@ -169,7 +176,7 @@ def self.run!(options = {}, &block)
         set :server, "thin"
         set :logger, Logger.new(STDOUT)
         set :job_queue, JobQueue.new()
-        set :db, APIDatabaseHelper.new("api.db")
+        set :db, SSHScan::Database::MongoDb.new
```
royalharsh: Yeah. I created a new database_config.yml .

lib/ssh_scan/api.rb
royalharsh: I have removed this. Will update the wiki too.

lib/ssh_scan/database/api_db_config.rb
```diff
@@ -0,0 +1,14 @@
...
+    def self.set_database
+      db_path = './lib/ssh_scan/database/database_config.yml'
+      db_opts = YAML.load_file(db_path)
+
+      if db_opts['database']['type'].eql? 'mongodb'
+        return SSHScan::Database::MongoDb.from_config_file(db_path)
+      elsif db_opts['database']['type'].eql? 'sqlite'
+        return SSHScan::Database::SQLite.from_config_file(db_path)
+      end
+    end
```
claudijd: Recommend replacing this file with a from_config method that hangs off of SSHScan::Database.

`db = SSHScan::Database.from_config(db_path)`

The method itself would be defined like so....

    module SSHScan
      class Database
        #rest of class
      
        def self.from_config(config_path)
         #some logic here that picks the DB type/object to pass as options
         SSHScan::Database.new(opts)
        end
      end
    end

lib/ssh_scan/database/database_config.yml
```diff
@@ -0,0 +1,20 @@
+# MongoDB Example
+database:
+  type: mongodb
+  name: api_db
```
claudijd: Would recommend "ssh_scan" for name here.

lib/ssh_scan/database/mongo_db.rb
```diff
@@ -0,0 +1,40 @@
+require 'mongo'
+
+Mongo::Logger.logger.level = ::Logger::FATAL
+
+module   SSHScan
+  module Database
+    class MongoDb
+      def initialize(opts = {})
+        name = opts["database"]["name"]
```
claudijd: We're making a lot of assumptions here about what might be in the config file.  I wonder if we should have some error handling here to make sure we have a minmum set of information to work with to avoid runtime errors.  Thoughts?

lib/ssh_scan/database/mongo_db.rb
```diff
@@ -0,0 +1,40 @@
...
+      end
+
+      def add_scan(worker_id, uuid, result)
+        @scans.insert_one("uuid" => uuid, "scan" => result,
+                          "worker_id" => worker_id, "scanned_on" => Time.now)
+      end
+
+      def delete_scan(uuid)
+        @scans.delete_one(:uuid => uuid)
+      end
```
claudijd: Just another comment here that in Ruby, this is what we're striving for, an Object class that has various interfaces, but the methods are really short, each to understand and document.  This makes for really maintainable less buggy code :)

lib/ssh_scan/database/mongo_db.rb
royalharsh: Yeah. I think we can include something like this, a default name?
```ruby
name = opts["database"]["name"] || "ssh_scan"
```

lib/ssh_scan/database/mongo_db.rb
```diff
@@ -0,0 +1,40 @@
+require 'mongo'
```
claudijd: The naming on these files is redundant.  I would consider removing the "_db" as you're already name spaced as a datbase in the file name and in the Ruby module namespace.

claudijd: Same goes for the other DB types.

lib/ssh_scan/database/sqlite_db.rb
```diff
@@ -0,0 +1,60 @@
...
+        @db.execute("delete from api_schema")
+      end
+
+      def find_scan_result(uuid)
+        scans = []
+        @db.execute(
+          "select * from api_schema where uuid like ( ? )",
+          uuid
+        ) do |row|
+          scans << row[1]
```
claudijd: For SQLite abstraction, we return all scans, but for Mongo we return the first scan.  I think we should pick a common behavior and make it consistent between the two.

ssh_scan.gemspec
```diff
@@ -29,6 +29,7 @@ Gem::Specification.new do |s|
   s.add_dependency('bindata', '~> 2.0')
   s.add_dependency('netaddr')
   s.add_dependency('net-ssh')
+  s.add_dependency('mongo')
```
claudijd: It's not a blocker just yet, but I'm starting to wonder how we'll want to deal with the converging dependencies that are being introduced by supporting multiple DB types.  I say this because we would want to avoid a situation where a user just wants ssh_scan binary, but they would need to install mongo/mysql/sqlite just to do that.

This is some food for thought, maybe we should discuss our options on the call.

lib/ssh_scan/database/sqlite_db.rb
royalharsh: Here also we are returning only one scan. Since, uuid is unique so we would get only a single scan.

royalharsh:
@claudijd updated.

lib/ssh_scan/api.rb
```diff
@@ -150,14 +152,7 @@ class API < Sinatra::Base
...
 
-        settings.db.find_work_result(worker_id, uuid)
-      end
-
-      get '/__version__' do
-        {
-          :ssh_scan_version => SSHScan::VERSION,
-          :api_version => SSHScan::API_VERSION,
-        }.to_json
+        settings.db.add_scan(worker_id, uuid, JSON.parse(request.body.first).first)
```
claudijd: If you ever task a multi-ip scan to a worker, this means that you will only ever record the first result of potentially many results that would be posted back.  I think this is ok if the expectation is that you'll only ever task a single target at a time via the API.

lib/ssh_scan/database.rb
```diff
@@ -0,0 +1,26 @@
+# API Databases
+require 'ssh_scan/database/mongo'
+require 'ssh_scan/database/sqlite'
+
+module SSHScan
+  class DatabaseConfig
+    def initialize(opts = {})
+      @@db_type = opts['database']['type'] || 'mongodb'
```
claudijd: You are using class variables here "@@" instead of instance variables "@".

Reference: https://www.tutorialspoint.com/ruby/ruby_variables.htm

I believe you want instance variables here.

lib/ssh_scan/database.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    def self.from_config_file
+      @@db_path = './lib/ssh_scan/database/database_config.yml'
```
claudijd: This value for db_path should not be hard coded.  It should be configurable.

lib/ssh_scan/database.rb
claudijd: You will also get NoMethod errors if the opts variable doesn't contain the 'database' key.  This is a blocking bug.

Example:

    2.2.3 :002 > opts = {}
     => {} 
    2.2.3 :003 > opts['database']['type']
    NoMethodError: undefined method `[]' for nil:NilClass
    	from (irb):3
    	from /Users/jclaudius/.rvm/rubies/ruby-2.2.3/bin/irb:15:in `<main>'

lib/ssh_scan/database.rb
claudijd: It should also use the expand path pattern to load it from a predictable location when it's bundled as a Gem.  Search for expand in the source tree and you'll find an example.

I'm also thinking that maybe we need a common location for configuration files.  Perhaps ./config and land them all there.

lib/ssh_scan/database/database_config.yml
```diff
@@ -0,0 +1,20 @@
+# MongoDB Example
+database:
+  type: mongodb
+  name: ssh_scan
+  server: 127.0.0.1
+  port: 27017
```
claudijd: As discussed in the MWoS meeting today, we should have SQLite as the default to make it easier for new users to stand up a basic implementation and if later they want to switch to mongo, all they need to do is change the config and restart their API server.

lib/ssh_scan/database/mongo.rb
```diff
@@ -0,0 +1,40 @@
+require 'mongo'
+
+Mongo::Logger.logger.level = ::Logger::FATAL
+
+module   SSHScan
+  module Database
+    class MongoDb
+      def initialize(opts = {})
+        name = opts["database"]["name"] || "ssh_scan"
```
claudijd: Same here, you will end up with NoMethod errors if the database key is not present.

lib/ssh_scan/database/sqlite.rb
```diff
@@ -0,0 +1,60 @@
+require 'sqlite3'
+
+module   SSHScan
+  module Database
+    class SQLite
+      def initialize(opts = {})
+        database_name = opts["database"]["name"] || "ssh_scan"
```
claudijd: NoMethod bug

spec/ssh_scan/api_spec.rb
```diff
@@ -20,7 +20,7 @@ def app
   end
 
   it "should be able to GET __version__ correctly" do
-    get "/api/v#{SSHScan::API_VERSION}/__version__"
+    get "/__version__"
```
claudijd: Good suggestion

lib/ssh_scan/database.rb
royalharsh: I tried instance variables at first. But, the `@db_type` wasn't getting populated with the type of database. So I used class variables here.

claudijd: It should definitely work with instance variables, I would try and again and if it doesn't work mention me and point me to the code and I can help you debug.

royalharsh: Fixed. :)

lib/ssh_scan/database.rb
royalharsh: Yeah, I think we should have a common place for config file. Something like, `./config/database/` for database config files and similarly for others.

claudijd: Sounds good, let's create an issue or just start doing it in this PR and ask others to follow.

lib/ssh_scan/database.rb
```diff
@@ -0,0 +1,30 @@
+# API Databases
+require 'ssh_scan/database/mongo'
+require 'ssh_scan/database/sqlite'
+
+module SSHScan
+  class DatabaseConfig
+    def initialize(opts = {})
+      @@db_type = if !opts || opts.empty?
```
royalharsh: On using `@db_type` instance variable here. In the `set_database` class method, `@db_type` is empty.

royalharsh: Last time , I tried: [Gist](https://gist.github.com/royalharsh/cc1066b6cf9faf9da570647e8af1f0db)

lib/ssh_scan/database.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    def self.from_config_file
+      @db_path = File.join(Dir.pwd, '/config/database/database_config.yml')
+      db_opts = YAML.load_file(@db_path)
+      SSHScan::DatabaseConfig.new(db_opts)
+    end
+
+    def self.set_database
+      from_config_file
+      if @@db_type.eql? 'mongodb'
```
royalharsh: `@db_type` would be empty here.

claudijd: This is because that is a class method and not an instance method.

http://www.railstips.org/blog/archives/2009/05/11/class-and-instance-methods-in-ruby/

claudijd: I think the best course of action here is to consolidate #from_config and #set_database methods.  I would also avoid the use of instance variables in class methods.

royalharsh: Yeah, I think consolidating methods might make things a bit easier.

lib/ssh_scan/database.rb
```diff
@@ -0,0 +1,26 @@
...
+  class DatabaseConfig
+    def initialize(opts = {})
+      @@db_type = if !opts || opts.empty?
+                   'sqlite'
+                 else
+                   opts['database']['type']
+                 end
+    end
+
+    def self.from_config_file
```
royalharsh: Squashed both the methods into one.

claudijd: @royalharsh I think all this file loading is getting overly confusing.

I'd like to propose we do something like this, with maybe even bringing the config loading out to the API level, since having DB configs and API configs is sort of odd behavior.  I've also mocked up an example on how we can make DB operations a little more transparent via a strong abstraction layer..

https://gist.github.com/claudijd/4004ad6c601d6b199bdf704e022857d2

claudijd: In the end, I think this will make it easier to follow and will hopefully add clarity around the complexity/confusion we're seeing around class/instance methods/variables

claudijd:
@royalharsh assuming what you have here works, I'd like to land this today and then I'd like to refactor a couple bits immediately after.  It's a little hard to send direct PRs suggestions against your fork.

claudijd:
Nice job @royalharsh and @rishabhs95, just landed.  I'll be sending a follow up PR with some tweaks, I'll cc you and @pwnbus for peer review on those updates.

royalharsh:
Sure!

Issue #293, opened 2016-12-14
Need banner-based conditional policy selection


claudijd:
If I'm running two different versions of OpenSSH, let's say 5.3 and 6.x, I have different capabilities available to me.  It would be cool if a policy to support some sort of conditional logic that could interpret the banner and apply alternative policies requirements for each.

Right now, the work around is that you would create a policy for 5.3 and create a policy for 6.x and then scan your 5.3 servers with the 5.3 policy and the same respectively for 6.x.  This is a bit clunky and we can probably rely on computers to make this distinction.

This almost feels like it could be a set of policies and we could do a first match sort of situation on fitting policy criteria and then have a default fallback policy.

agaurav77:
So, does this mean we'll have more yaml files in `policies`, and then if results are OpenSSH, then we apply a policy file based on the OpenSSH version (and have a fallback, say `mozilla_modern.yml`)?

agaurav77:
This will have to be two-phase
(1) land in code to match `policies/openssh_version.yml`
(2) land in multiple yml files

claudijd:
I could even see having something like this...

policies/openssh/6.0.yml
policies/openssh/5.3.yml
policies/openssh/default.yml

And then just load them into the policy engine in the order they should apply.

claudijd:
@agaurav77 and I spoke about this a bit more via Glitter.  we talked about having a rule/requirements qualifier to determine if a set of policies is relevant for a given policy application.

Example: https://gist.github.com/claudijd/d34df76e0248e190b5a9b7b4412db28d

agaurav77:
Based on @claudijd's input, I drafted an initial page on [policy expression DSL](https://github.com/mozilla/ssh_scan/wiki/Policy-Expression-DSL). Ideally I would love to go towards to the possible extensions to allow maximum flexibility for the end user, but we would work on getting the original conditional based policy expression working first.

gerhard-tinned:
Nice idea. So we are talking about sime kind of policy group thing. 

What about this, ...

A policy like file, lets call it condittion file, containing a list of ... lets call it conditions. One of the configuration in the condition could be a reference to another policy file. 

It could be like this (based on the draft):

<pre>
Rulesets:
  - rules:
  or:
    and:
      ssh_version_eq: 2.0
      ssh_lib_eq: openssh
    and:
      ssh_version_eq: 1.34
      ssh_lib_eq: dropbear
    policy: subdirectory/mozilla_modern.yml
[...]

PR #295, opened 2016-12-15
add rule_engine, first match policy coverage

ashishgaurav13:
for #293

### DIFF
--- a/bin/ssh_scan
+++ b/bin/ssh_scan
@@ -12,7 +12,6 @@ require 'logger'
 #Default options
 options = {
   "sockets" => [],
-  "policy" => File.join(Dir.pwd, '/config/policies/mozilla_modern.yml'),
   "unit_test" => false,
   "timeout" => 2,
   "threads" => 5,
@@ -127,6 +126,12 @@ scan") do |file|
   opts.on("-P", "--policy [FILE]",
           "Custom policy file (Default: Mozilla Modern)") do |policy|
     options["policy"] = policy
+    unless File.exist?(options["policy"])
+      puts opt_parser.help
+      puts "\nReason: policy file supplied is not a file"
+      exit 1
+    end
+    options["policy_file"] = SSHScan::Policy.from_file(options["policy"])
   end
 
   opts.on("--threads [NUMBER]",
@@ -226,12 +231,6 @@ options["sockets"].each do |socket|
   end
 end
 
-unless File.exist?(options["policy"])
-  puts opt_parser.help
-  puts "\nReason: policy file supplied is not a file"
-  exit 1
-end
-
 # Check to see if we're running the latest released version
 if !options["suppress_update_status"]
   update = SSHScan::Update.new
@@ -252,8 +251,6 @@ ssh_scan' to get the latest"
     end
   end
 end
-
-options["policy_file"] = SSHScan::Policy.from_file(options["policy"])
 
 # Perform scan and get results
 scan_engine = SSHScan::ScanEngine.new()
--- a/lib/ssh_scan/policy.rb
+++ b/lib/ssh_scan/policy.rb
@@ -1,4 +1,5 @@
 require 'yaml'
+require 'logger'
 
 module SSHScan
   class Policy
@@ -14,6 +15,7 @@ module SSHScan
       @references = opts['references'] || []
       @auth_methods = opts['auth_methods'] || []
       @ssh_version = opts['ssh_version'] || false
+      @logger = opts['logger']
     end
 
     def self.from_file(file)
@@ -24,6 +26,25 @@ module SSHScan
     def self.from_string(string)
       opts = YAML.load(string)
       self.new(opts)
+    end
+
+    # Choose a policy file based on rules
+    def self.first_match(result)
+      logger = @logger || Logger.new(STDERR)
+      policies_dir = File.join(File.dirname(__FILE__), "../../config/policies")
+      fallback_file = File.join(policies_dir, "mozilla_modern.yml")
+      Dir.glob(policies_dir + "/**/*.yml") do |policy_file|
+        next if policy_file == fallback_file
+        file = YAML.load_file(policy_file)
+        next unless file["rules"]
+        logger.info("Matching: #{policy_file}")
+        if SSHScan::RuleEngine.matchRuleset(file["rules"], file["ruleset"], result)
+          logger.info("Matched: #{policy_file}")
+          return policy_file
+        end
+      end
+      logger.info("No matches found, using fallback policy file (#{fallback_file}).")
+      return fallback_file
     end
   end
 end
--- a/lib/ssh_scan/rule_engine.rb
+++ b/lib/ssh_scan/rule_engine.rb
@@ -1,0 +1,99 @@
+module SSHScan
+  class RuleEngine
+    # Simple one rule matching.
+    def self.matchSingleRule(key, value, result)
+      conditional = key.split("_")[-1]
+      attribute = key.chomp("_" + conditional).to_sym
+      rvalue = result[attribute]
+      if rvalue
+        answer = case conditional
+          when "eq" then (rvalue == value)
+          when "neq" then (rvalue != value)
+        end
+        return answer if answer
+      end
+      if rvalue.is_a? Numeric
+        answer = case conditional
+          when "gt" then (rvalue > value)
+          when "lt" then (rvalue < value)
+          when "gte" then (rvalue >= value)
+          when "lte" then (rvalue <= value)
+        end
+        return answer if answer
+      end
+      return false
+    end
+
+    # Match a simple ruleset condition using the stack based parser.
+    def self.matchRuleset(rules, ruleset, result)
+      mem = {}
+      rules.each do |rule_name, values|
+        mem[rule_name] = matchSingleRule(values[0], values[1], result)
+      end
+      return parseBooleanRule(ruleset, mem)
+    end
+
+    # Check if o is boolean or not. Apparently is_a?(Boolean) is not a thing.
+    def self.checkBool(o)
+      !!o == o
+    end
+
+    # Reduce boolean expression parser stack once.
+    # Examples of stacks and what reduction does -
+    # ( T && ( F && T   =>    ( T && ( F
+    # ( T || ( F || T   =>    ( T || ( T
+    # ( T && ! F        =>    ( T && T
+    def self.reduceStack!(stack)
+      if (stack[-2] == '!') && checkBool(stack[-1])
+        new_tokenval = !stack[-1]
+        2.times { stack.pop }
+        stack << new_tokenval
+      elsif (stack[-2] == '&&') && checkBool(stack[-3]) && checkBool(stack[-1])
+        new_tokenval = stack[-1] && stack[-3]
+        3.times { stack.pop }
+        stack << new_tokenval
+      elsif (stack[-2] == '||') && checkBool(stack[-3]) && checkBool(stack[-1])
+        new_tokenval = stack[-1] || stack[-3]
+        3.times { stack.pop }
+        stack << new_tokenval
+      else
+        # if we cannot reduce, say so
+        return false
+      end
+      # if we exited from the if, we have reduced something, so say so
+      return true
+    end
+
+    # Parse a boolean expression with parentheses, uses stack.
+    def self.parseBooleanRule(rule, inputs)
+      tokens = rule.delete(' ').gsub('(', ' ( ').gsub(')', ' ) ')
+                   .gsub('&&', ' && ').gsub('||', ' || ').gsub('!', ' ! ').split
+      stack = []
+      tokens.each do |token|
+        case token
+          when /^[a-zA-Z0-9]*$/
+            tokenval = if inputs.include?(token) then inputs[token]
+                       elsif inputs.include?(token.to_sym) then inputs[token.to_sym]
+              end
+            return nil if tokenval.nil?
+            stack << tokenval
+          when "(", "&&", "||", "!"
+            stack << token
+          when ")"
+            if stack[-2] == "("
+              new_val = stack[-1]
+              2.times { stack.pop }
+              stack << new_val
+            else
+              return nil
+            end
+        end
+        loop do
+          break unless reduceStack!(stack)
+        end
+      end
+      return stack[0] if stack.length == 1
+      return nil
+    end
+  end
+end
--- a/spec/ssh_scan/rule_engine_spec.rb
+++ b/spec/ssh_scan/rule_engine_spec.rb
@@ -1,0 +1,150 @@
+require 'rspec'
+require 'ssh_scan/rule_engine'
+
+# ignore absurdities in rules, they have been created to test the rule engine
+describe SSHScan::RuleEngine do
+  context "match single rule" do
+    it "eq should work for non numerics" do
+      result = {:ssh_lib => "openssh"}
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_lib_eq", "openssh", result)).to be true
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_lib_eq", "libssh", result)).to be false
+    end
+
+    it "neq should work for non numerics" do
+      result = {:ssh_lib => "openssh"}
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_lib_neq", "dropbear", result)).to be true
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_lib_neq", "openssh", result)).to be false
+    end
+
+    it "eq should work for numerics" do
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_eq", 2.0, result)).to be true
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_eq", 1.99, result)).to be false
+    end
+
+    it "neq should work for numerics" do
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_neq", 1.99, result)).to be true
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_neq", 2.0, result)).to be false
+    end
+
+    it "gt should work for numerics" do
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_gt", 1.99, result)).to be true
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_gt", 2.0, result)).to be false
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_gt", 2.1, result)).to be false
+    end
+
+    it "gte should work for numerics" do
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_gte", 1.99, result)).to be true
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_gte", 2.0, result)).to be true
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_gte", 2.1, result)).to be false
+    end
+
+    it "lt should work for numerics" do
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_lt", 1.99, result)).to be false
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_lt", 2.0, result)).to be false
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_lt", 2.1, result)).to be true
+    end
+
+    it "lte should work for numerics" do
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_lte", 2.1, result)).to be true
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_lte", 2.0, result)).to be true
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_lte", 1.99, result)).to be false
+    end
+
+    it "should not ignore attributes not in results" do
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_lib_eq", "openssh", result)).to be false
+    end
+  end
+
+  # Note: inputs can be given in a hash as well
+  context "parse conditional boolean expressions" do
+    it "simple true/false" do
+      expect(SSHScan::RuleEngine.parseBooleanRule("x", "x" => true)).to be true
+      expect(SSHScan::RuleEngine.parseBooleanRule("x", "x" => false)).to be false
+    end
+
+    it "should work for NOT expressions" do
+      expect(SSHScan::RuleEngine.parseBooleanRule("!x", "x" => true)).to be false
+      expect(SSHScan::RuleEngine.parseBooleanRule("!x", "x" => false)).to be true
+    end
+
+    it "should work for AND expressions" do
+      expect(SSHScan::RuleEngine.parseBooleanRule("x && y", "x" => true, "y" => true)).to be true
+      expect(SSHScan::RuleEngine.parseBooleanRule("x && y", "x" => true, "y" => false)).to be false
+      expect(SSHScan::RuleEngine.parseBooleanRule("x && y", "x" => false, "y" => true)).to be false
+      expect(SSHScan::RuleEngine.parseBooleanRule("x && y", "x" => false, "y" => false)).to be false
+    end
+
+    it "should work for OR expressions" do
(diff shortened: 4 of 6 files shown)


lib/ssh_scan/policy.rb
```diff
@@ -25,5 +26,20 @@ def self.from_string(string)
       opts = YAML.load(string)
       self.new(opts)
     end
+
+    # Choose a policy file based on rules
+    def self.first_match(result)
+      Dir.glob("policies/**/*.yml") do |policy_file|
```
claudijd: This might not work once we release the code as a gem, consider expanding the path based on current location of this file.

lib/ssh_scan/policy.rb
```diff
@@ -25,5 +26,20 @@ def self.from_string(string)
...
+        next if policy_file == @fallback_file
+        file = YAML.load_file(policy_file)
+        next unless file["rules"]
+        puts "Matching: #{policy_file}"
+        if SSHScan::RuleEngine.match(file["rules"], result)
+          puts "Matched: #{policy_file}"
```
claudijd: Same here, RE: logger

lib/ssh_scan/policy.rb
agaurav77: Since ssh_scan first chdirs to the root directory, and it is the only script calling this function, it should work, right?

lib/ssh_scan/policy.rb
```diff
@@ -25,5 +26,20 @@ def self.from_string(string)
...
+          policy_file
```
claudijd: do you want "return" before this?

agaurav77: I can do that to make it clearer, but essentially return ... will do the same thing, right?

lib/ssh_scan/policy.rb
```diff
@@ -14,6 +14,7 @@ def initialize(opts = {})
       @references = opts['references'] || []
       @auth_methods = opts['auth_methods'] || []
       @ssh_version = opts['ssh_version'] || false
+      @fallback_file = "policies/mozilla_modern.yml"
```
claudijd: Same here, with the file reference.

lib/ssh_scan/rule_engine.rb
```diff
@@ -0,0 +1,30 @@
...
+              answer &= (result[attribute] > value)
+            elsif conditional == "lt"
+              answer &= (result[attribute] < value)
+            elsif conditional == "gte"
+              answer &= (result[attribute] >= value)
+            elsif conditional == "lte"
+              answer &= (result[attribute] <= value)
+            end
+          end
+        end
```
claudijd: This block of nested if's might be a better fit as a case statement or it's own unique object class that can cleanly parse a single rule.

spec/ssh_scan/rule_engine_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'rspec'
+require 'ssh_scan/rule_engine'
+
+describe SSHScan::RuleEngine do
+  context "for conditionals" do
+    it "eq, neq should work for any literal" do
+      result = {:ssh_lib => "openssh"}
+      expect(SSHScan::RuleEngine.match({"ssh_lib_eq" => "openssh"}, result)).to be_truthy
```
claudijd: recommend be_true here.   In ruby, "true" is not the same as "truthy".

Since your match routine is always true/false, I would recommend constraining your unit-tests to look for those specifically.

Reference: for more truthy education (https://gist.github.com/jfarmer/2647362)

agaurav77: Will change, did not know about this one. Thanks!

lib/ssh_scan/policy.rb
claudijd: In ruby, you get an implicit return if it's the last operation in the scope.

def example
   foo #will return foo, but implicitly
end

def example
   return foo #will also return foo, but explicitly
end

but if you want an early return, you need to be explicit...

def example
  if foo
    return bar #if foo, then return bar
  end

  foo # if not foo, then default to implicit return of foo
end

lib/ssh_scan/policy.rb
claudijd: My understanding would be that it will work when you run from source, but not when you run as a gem.

You can test this by doing the following:

1.) gem build ssh_scan.gemspec # build a gem
2.) gem install ssh_scan-0.0.16.gem # install the gem
3.) go to a different directory and run to exercise this code in IRB and see if the path still works

lib/ssh_scan/policy.rb
agaurav77: Didn't know this difference, thanks for pointing out! Changing..

agaurav77:
@claudijd please re review

spec/ssh_scan/rule_engine_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'rspec'
+require 'ssh_scan/rule_engine'
+
+describe SSHScan::RuleEngine do
+  context "for conditionals" do
+    it "eq, neq should work for any literal" do
+      result = {:ssh_lib => "openssh"}
+      expect(SSHScan::RuleEngine.match({"ssh_lib_eq" => "openssh"}, result)).to be true
+      expect(SSHScan::RuleEngine.match({"ssh_lib_neq" => "dropbear"}, result)).to be true
```
claudijd: I would recommend adding this test as well...

`expect(SSHScan::RuleEngine.match({"ssh_lib_eq" => "dropbear"}, result)).to be false`

spec/ssh_scan/rule_engine_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.match({"ssh_version_eq" => 1.99}, result)).to be false
+      expect(SSHScan::RuleEngine.match({"ssh_version_neq" => 1.99}, result)).to be true
```
claudijd: I would recommend adding this test as well...

`expect(SSHScan::RuleEngine.match({"ssh_version_eq" => 2.0}, result)).to be true`

spec/ssh_scan/rule_engine_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    it "gt, lt, gte, lte should work for numeric literal" do
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.match({"ssh_version_gt" => 1.99}, result)).to be true
+      expect(SSHScan::RuleEngine.match({"ssh_version_gte" => 2.0}, result)).to be true
+      expect(SSHScan::RuleEngine.match({"ssh_version_lt" => 1.99}, result)).to be false
+      expect(SSHScan::RuleEngine.match({"ssh_version_lte" => 2.0}, result)).to be true
```
claudijd: Recommend testing below/eql/above examples (to verify boundaries) for each verb, like so...

    # Testing gt boundaries
    expect(SSHScan::RuleEngine.match({"ssh_version_gt" => 1.99}, result)).to be true
    expect(SSHScan::RuleEngine.match({"ssh_version_gt" => 2.0}, result)).to be false
    expect(SSHScan::RuleEngine.match({"ssh_version_gt" => 2.1}, result)).to be false

    # Testing gte boundaries
    expect(SSHScan::RuleEngine.match({"ssh_version_gte" => 1.99}, result)).to be true
    expect(SSHScan::RuleEngine.match({"ssh_version_gte" => 2.0}, result)).to be true
    expect(SSHScan::RuleEngine.match({"ssh_version_gte" => 2.1}, result)).to be false

    # Testing lt boundaries
    expect(SSHScan::RuleEngine.match({"ssh_version_lt" => 1.99}, result)).to be false
    expect(SSHScan::RuleEngine.match({"ssh_version_lt" => 2.0}, result)).to be false
    expect(SSHScan::RuleEngine.match({"ssh_version_lt" => 2.1}, result)).to be true

    # Testing lte boundaries
    expect(SSHScan::RuleEngine.match({"ssh_version_lte" => 1.99}, result)).to be false
    expect(SSHScan::RuleEngine.match({"ssh_version_lte" => 2.0}, result)).to be true
    expect(SSHScan::RuleEngine.match({"ssh_version_lte" => 2.1}, result)).to be true

spec/ssh_scan/rule_engine_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    it "should ignore attributes not in results" do
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.match({"ssh_lib_eq" => "openssh"}, result)).to be true
```
claudijd: Shouldn't the test expectation here be "false" and not "true"?

claudijd: My expectation here is that you need to have two elements that match before a match would be "true", but the example case is matching true to two different elements.

agaurav77:
@claudijd 

1. Most of the online boolean expression parsers were either broken or overkill for simple expression parsing, so I created a simple one using an array-based stack.

2. I have added sufficient tests to check if the expression parsing is okay, feel free to play around with the tests.

3. It is very easy to see how the policy file first match would work (like in the updated [wiki](https://github.com/mozilla/ssh_scan/wiki/Policy-Expression-DSL)). Just create a file in the `policies` folder (copy it over from say `mozilla_intermediate.yml`). Add in a few conditions like

    ```yaml
    rules:
      a:
        - ssh_lib_eq
        - openssh
      b:
        - ssh_version_gt
        - 1.99
    ruleset: "!a && ((b))" # parentheses supported
    ...
    ```
    When you run it on an endpoint, it should try to match against this file (and will show it in the logs, stdout/stderr).

    I am going with this notation for `rules` because if I instead have something like `"ssh_lib_eq openssh"`, I have to additionally parse the value. YAML conversion handles it if I do it this way.

I hope this should be sufficient for #293. Also do checkout the checklist in the updated [wiki](https://github.com/mozilla/ssh_scan/wiki/Policy-Expression-DSL)!

bin/ssh_scan
```diff
@@ -127,6 +126,12 @@ scan") do |file|
   opts.on("-P", "--policy [FILE]",
           "Custom policy file (Default: Mozilla Modern)") do |policy|
     options["policy"] = policy
+    unless File.exist?(options["policy"])
+      puts opt_parser.help
+      puts "\nReason: policy file supplied is not a file"
+      exit 1
+    end
+    options["policy_file"] = SSHScan::Policy.from_file(options["policy"])
```
claudijd: This is a behavior change in which a user will always be required to specify a policy file.  Is this truly the behavior we're shooting for?

lib/ssh_scan/policy.rb
```diff
@@ -25,5 +27,23 @@ def self.from_string(string)
...
+      Dir.glob(policies_dir + "/**/*.yml") do |policy_file|
+        next if policy_file == fallback_file
+        file = YAML.load_file(policy_file)
+        next unless file["rules"]
+        logger.info("Matching: #{policy_file}")
+        if SSHScan::RuleEngine.matchRuleset(file["rules"], file["ruleset"], result)
+          logger.info("Matched: #{policy_file}")
+          return policy_file
+        end
+      end
```
claudijd: Maybe we should log that we ended up using the fall back?

lib/ssh_scan/rule_engine.rb
```diff
@@ -0,0 +1,94 @@
+module SSHScan
+  class RuleEngine
+    # Simple one rule matching.
+    def self.matchSingleRule(key, value, result)
```
claudijd: Maybe we should start using YARD documentation for these interfaces, so it's more clear what each parameter class type is?

lib/ssh_scan/rule_engine.rb
```diff
@@ -0,0 +1,94 @@
...
+      answer = true
```
claudijd: Should the default answer be false and not true?

I'd rather fall back than incorrectly match.

lib/ssh_scan/rule_engine.rb
```diff
@@ -0,0 +1,94 @@
...
+          when "gte" then (rvalue >= value)
+          when "lte" then (rvalue <= value)
+          else true
+        end
+      end
+      answer
+    end
+
+    # Match a simple ruleset condition using the stack based parser.
+    def self.matchRuleset(rules, ruleset, result)
```
claudijd: Again, here is a good place where YARD docs would help

lib/ssh_scan/rule_engine.rb
```diff
@@ -0,0 +1,94 @@
...
+      mem = {}
+      rules.each do |rule_name, values|
+        mem[rule_name] = matchSingleRule(values[0], values[1], result)
+      end
+      return parseBooleanRule(ruleset, mem)
+    end
+
+    # Check if o is boolean or not. Apparently is_a?(Boolean) is not a thing.
```
claudijd: yeah, in Ruby, you've got True class and False classes.

lib/ssh_scan/rule_engine.rb
```diff
@@ -0,0 +1,94 @@
...
+        stack << new_tokenval
+      elsif (stack[-2] == '||') && checkBool(stack[-3]) && checkBool(stack[-1])
+        new_tokenval = stack[-1] || stack[-3]
+        3.times { stack.pop }
+        stack << new_tokenval
+      else
+        return false
+      end
+      return true
+    end
```
claudijd: I'm not sure what this is doing. Maybe some more documentation around this functionality would help.

spec/ssh_scan/rule_engine_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+    it "neq should work for numerics" do
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_neq", 1.99, result)).to be true
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_neq", 2.0, result)).to be false
+    end
+
+    it "gt should work for numerics" do
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_gt", 1.3, result)).to be true
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_gt", 2.0, result)).to be false
```
claudijd: Recommend using the following cases:

expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_gt", 1.99, result)).to be true
expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_gt", 2.0, result)).to be false
expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_gt", 2.1, result)).to be false

claudijd: Recommending 1.99 because it's common and a closer boundary to check.

Recommending 2.1 because it tests the other side of the boundary.

spec/ssh_scan/rule_engine_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+    end
+
+    it "gte should work for numerics" do
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_gte", 1.3, result)).to be true
```
claudijd: Recommend replacing 1.3 with...

expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_gt", 1.99, result)).to be true

spec/ssh_scan/rule_engine_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_lt", 2.1, result)).to be true
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_lt", 2.0, result)).to be false
+    end
+
+    it "lte should work for numerics" do
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_lte", 2.1, result)).to be true
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_lte", 2.0, result)).to be true
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_lte", 1.3, result)).to be false
```
claudijd: Recommend replacing...

expect(SSHScan::RuleEngine.matchSingleRule("ssh_version_lte", 1.99, result)).to be false

spec/ssh_scan/rule_engine_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+    end
+
+    it "should ignore attributes not in results" do
+      result = {:ssh_version => 2.0}
+      expect(SSHScan::RuleEngine.matchSingleRule("ssh_lib_eq", "openssh", result)).to be true
```
claudijd: I would expect this to fail, as the default behavior when the key requirement is not present should be to return false, right?

  expect(SSHScan::RuleEngine.matchSingleRule("ssh_lib_eq", "openssh", result)).to be false

spec/ssh_scan/rule_engine_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+    it "should work for NOT expressions" do
+      expect(SSHScan::RuleEngine.parseBooleanRule("!x", "x" => true)).to be false
+      expect(SSHScan::RuleEngine.parseBooleanRule("!x", "x" => false)).to be true
+    end
+
+    it "should work for AND expressions" do
+      expect(SSHScan::RuleEngine.parseBooleanRule("x && y", "x" => true, "y" => true)).to be true
+      expect(SSHScan::RuleEngine.parseBooleanRule("x && y", "x" => true, "y" => false)).to be false
+      expect(SSHScan::RuleEngine.parseBooleanRule("x && y", "x" => false, "y" => true)).to be false
+      expect(SSHScan::RuleEngine.parseBooleanRule("x && y", "x" => false, "y" => false)).to be false
```
claudijd: I like that you're going through all the cases here, having complete coverage can give you a lot more confidence in the code not having logic/boundary bugs.

spec/ssh_scan/rule_engine_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+    it "should work for OR expressions" do
+      expect(SSHScan::RuleEngine.parseBooleanRule("x || y", "x" => true, "y" => true)).to be true
+      expect(SSHScan::RuleEngine.parseBooleanRule("x || y", "x" => true, "y" => false)).to be true
+      expect(SSHScan::RuleEngine.parseBooleanRule("x || y", "x" => false, "y" => true)).to be true
+      expect(SSHScan::RuleEngine.parseBooleanRule("x || y", "x" => false, "y" => false)).to be false
+    end
+
+    it "should work for compounded non bracketed AND, OR, NOT expressions" do
+      expect(SSHScan::RuleEngine.parseBooleanRule("!x && y", "x" => false, "y" => true)).to be true
+      expect(SSHScan::RuleEngine.parseBooleanRule("x || !y", "x" => false, "y" => false)).to be true
```
claudijd: Would recommend expanding the test cases here to include....
 
    expect(SSHScan::RuleEngine.parseBooleanRule("x || !y", "x" => false, "y" => false)).to be true
    expect(SSHScan::RuleEngine.parseBooleanRule("!x || y", "x" => false, "y" => false)).to be true
    expect(SSHScan::RuleEngine.parseBooleanRule("x || y", "x" => false, "y" => false)).to be false
    expect(SSHScan::RuleEngine.parseBooleanRule("!x || !y", "x" => false, "y" => false)).to be true

    expect(SSHScan::RuleEngine.parseBooleanRule("x || !y", "x" => false, "y" => true)).to be false
    expect(SSHScan::RuleEngine.parseBooleanRule("!x || y", "x" => false, "y" => true)).to be true
    expect(SSHScan::RuleEngine.parseBooleanRule("x || y", "x" => false, "y" => true)).to be true
    expect(SSHScan::RuleEngine.parseBooleanRule("!x || !y", "x" => false, "y" => true)).to be true

    expect(SSHScan::RuleEngine.parseBooleanRule("x && !y", "x" => false, "y" => false)).to be false
    expect(SSHScan::RuleEngine.parseBooleanRule("!x && y", "x" => false, "y" => false)).to be false
    expect(SSHScan::RuleEngine.parseBooleanRule("x && y", "x" => false, "y" => false)).to be false
    expect(SSHScan::RuleEngine.parseBooleanRule("!x && !y", "x" => false, "y" => false)).to be true

    expect(SSHScan::RuleEngine.parseBooleanRule("x && !y", "x" => false, "y" => true)).to be false
    expect(SSHScan::RuleEngine.parseBooleanRule("!x && y", "x" => false, "y" => true)).to be true
    expect(SSHScan::RuleEngine.parseBooleanRule("x && y", "x" => false, "y" => true)).to be false
    expect(SSHScan::RuleEngine.parseBooleanRule("!x && !y", "x" => false, "y" => true)).to be false

spec/ssh_scan/rule_engine_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+    end
+
+    it "should work for compounded bracketed AND, OR, NOT expressions" do
+      expect(SSHScan::RuleEngine.parseBooleanRule("x && (y || z)",
+        "x" => false, "y" => true, "z" => false)).to be false
```
claudijd: Please add...

    expect(SSHScan::RuleEngine.parseBooleanRule("x && (y || z)",  "x" => false, "y" => true, "z" => true)).to be true

spec/ssh_scan/rule_engine_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+      result2 = {:ssh_version => 1.99, :ssh_lib => "dropbear"}
+      result3 = {:ssh_version => 2.1, :ssh_lib => "dropbear"}
+      rules = {
+        "a" => ["ssh_version_gt", 1.99],
+        "b" => ["ssh_lib_eq", "openssh"],
+        "c" => ["ssh_lib_eq", "dropbear"],
+        "d" => ["ssh_version_lt", 1.99]
+      }
+      ruleset = "(a && b) || (c && !d)" # !d means ssh_version_gte 1.99
+      expect(SSHScan::RuleEngine.matchRuleset(rules, ruleset, result1)).to be true
```
claudijd: Walking through the logic out loud...

- ssh version 2.0 is gt than 1.99 (OK)
- ssh lib openssh is eq openssh (OK)

Overall: OK

spec/ssh_scan/rule_engine_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+      expect(SSHScan::RuleEngine.matchRuleset(rules, ruleset, result2)).to be true
```
claudijd: Walking through the logic out loud...

    ssh version 1.99 is NOT gt than 1.99 (OK)
    ssh lib dropbear is eq dropbear (OK)
    ssh version 1.99 is gte than 1.99 (OK)

Overall: OK

spec/ssh_scan/rule_engine_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+      expect(SSHScan::RuleEngine.matchRuleset(rules, ruleset, result3)).to be true
```
claudijd: Walking through the logic out loud...

    ssh version 2.1 is gt than 1.99 (OK)
    ssh lib dropbear is NOT eq openssh (OK)
    ssh lib dropbear is eq dropbear (OK)
    ssh version 2.1 is gte than 1.99 (OK)

Overall: OK

claudijd:
Overall, looks pretty good.  I added some suggestions about additional/modifications for test cases.  I also think that we should reconsider how we treat matches when the element is not present in the result.  The boolean/stack ranking stuff is a little hard to follow as a ruby person, but you've a good foundation for testing that leads me to believe it works as expected.

bin/ssh_scan
agaurav77: From what I understand, the user doesn't need to always specify a policy file. But when he/she doesn't, the policy file is found on a first match basis, failing which we use the fallback file.

lib/ssh_scan/rule_engine.rb
agaurav77: This definitely looks promising. Opened up #319 for this. I think we should pursue using YARD in a different PR.

lib/ssh_scan/rule_engine.rb
agaurav77: I agree. I was looking from an AND perspective, but it is certainly possible to do this. Will change.

lib/ssh_scan/rule_engine.rb
agaurav77: Added some examples, comments.

spec/ssh_scan/rule_engine_spec.rb
agaurav77: Yeah, according to default behavior as you suggested above, it should. Changing...

spec/ssh_scan/rule_engine_spec.rb
agaurav77: Shouldn't this be false?

agaurav77:
@claudijd should be good now, please take a look!

claudijd:
Blocked by #307

Issue #296, opened 2016-12-15
Provide a caching feature in the API to retrieve pre-existing scan result rather than re-task a worker


claudijd:
The basic problem statement is what User A and User B request the same scan within say 15min of each other.  How do we avoid doing the work twice?

**Solution:** provide a caching feature in the API that will retrieve a pre-existing scan from the database rather than handing this work off to a worker.

rishabhs95:
The basic caching functionality was included in PR#305. As the DB functionality is landed successfully, I am working to complete the missing parts of the functionality.

Currently, caching functionality just looks for the already present records and if a record is present (i.e, target is already scanned), we return its `uuid` instead of sending a new job to the worker. I think that we can move forward with these steps:

- We let the above functionality remain same.
- Additionally, we now also remove the already present scan in our DB for a target which is forced for a re-scan. So now, whenever the user requests the same target, they will receive the latest scan results.

I think that this should suffice our need for now. Do you think that we should also check for timestamps, for example, if the present scan in the DB is more than 2 days ago, initiate a re-scan and replace its result with the older one? @claudijd @agaurav77 @royalharsh

royalharsh:
Yeah, I think we should implement it. So that users get the latest scan.

rishabhs95:
Yeah, I was equally tending towards both the ways. I have included the timeouts for 1 day anyways in my PR.

claudijd:
This was done a long time ago, closing.

PR #305, opened 2016-12-22
API caching feature

rishabhs95:
Provides a caching feature in the API retrieving pre-existing scan results when present.

I am still working on it but opening this PR so that everyone can see where it is headed and suggest improvements.

This PR is separate from but is based on changes from PR #291, so it contains changes from earlier PR mixed with this too.
This addresses Issue #296.

### DIFF
--- a/lib/ssh_scan/api.rb
+++ b/lib/ssh_scan/api.rb
@@ -166,13 +166,15 @@ https://github.com/mozilla/ssh_scan/wiki/ssh_scan-Web-API\n"
         uuid = params['uuid']
         result = JSON.parse(request.body.first).first
         socket = {}
-        socket[:ip] = result['ip']
+        socket[:target] = result['ip']
         socket[:port] = result['port']
 
         if worker_id.empty? || uuid.empty?
           return {"accepted" => "false"}.to_json
         end
 
+        available_result = settings.db.fetch_available_result(socket)
+        settings.db.delete_scan(available_result) unless available_result.nil?
         settings.db.add_scan(worker_id, uuid, result, socket)
       end
 
--- a/lib/ssh_scan/api.rb
+++ b/lib/ssh_scan/api.rb
@@ -107,12 +107,18 @@ https://github.com/mozilla/ssh_scan/wiki/ssh_scan-Web-API\n"
         options[:force] = params[:force] ? params[:force] : false
         unless (options[:force] == 'true')
           available_result = settings.db.fetch_available_result(params)
-          return { uuid: available_result }.to_json unless available_result.nil?
+          unless available_result.nil? || ((Time.now - available_result[:scanned_on])/(60*60*24) > 1)
+            return {
+              uuid: available_result[:uuid],
+              scanned_on: available_result[:scanned_on].localtime
+            }.to_json
+          end
         end
         options[:uuid] = SecureRandom.uuid
         settings.job_queue.add(options)
         {
-          uuid: options[:uuid]
+          uuid: options[:uuid],
+          scanned_on: 'fresh scan'
         }.to_json
       end
 
@@ -174,7 +180,7 @@ https://github.com/mozilla/ssh_scan/wiki/ssh_scan-Web-API\n"
         end
 
         available_result = settings.db.fetch_available_result(socket)
-        settings.db.delete_scan(available_result) unless available_result.nil?
+        settings.db.delete_scan(available_result[:uuid]) unless available_result.nil?
         settings.db.add_scan(worker_id, uuid, result, socket)
       end
 
--- a/lib/ssh_scan/api.rb
+++ b/lib/ssh_scan/api.rb
@@ -103,6 +103,11 @@ https://github.com/mozilla/ssh_scan/wiki/ssh_scan-Web-API\n"
         options[:sockets] <<
           "#{params[:target]}:#{params[:port] ? params[:port] : "22"}"
         options[:policy_file] = options[:policy]
+        options[:force] = params[:force] ? params[:force] : false
+        unless (options[:force] == 'true')
+          available_result = settings.db.fetch_available_result(params)
+          return { uuid: available_result }.to_json unless available_result.nil?
+        end
         options[:uuid] = SecureRandom.uuid
         settings.job_queue.add(options)
         {
@@ -158,12 +163,16 @@ https://github.com/mozilla/ssh_scan/wiki/ssh_scan-Web-API\n"
       post '/work/results/:worker_id/:uuid' do
         worker_id = params['worker_id']
         uuid = params['uuid']
+        result = JSON.parse(request.body.first).first
+        socket = {}
+        socket[:ip] = result['ip']
+        socket[:port] = result['port']
 
         if worker_id.empty? || uuid.empty?
           return {"accepted" => "false"}.to_json
         end
 
-        settings.db.add_scan(worker_id, uuid, JSON.parse(request.body.first).first)
+        settings.db.add_scan(worker_id, uuid, result, socket)
       end
 
       get '/__lbheartbeat__' do
--- a/lib/ssh_scan/api.rb
+++ b/lib/ssh_scan/api.rb
@@ -18,7 +18,7 @@ module SSHScan
       configure do
         set :job_queue, JobQueue.new()
         set :authentication, false
-        set :db, SSHScan::DatabaseConfig.from_config_file
+        set :db, SSHScan::Database.from_hash(options)
       end
     end
 
--- a/lib/ssh_scan/api.rb
+++ b/lib/ssh_scan/api.rb
@@ -18,6 +18,7 @@ module SSHScan
       configure do
         set :job_queue, JobQueue.new()
         set :authentication, false
+        set :db, SSHScan::DatabaseConfig.from_config_file
       end
     end
 
--- a/lib/ssh_scan/api.rb
+++ b/lib/ssh_scan/api.rb
@@ -18,10 +18,7 @@ module SSHScan
       configure do
         set :job_queue, JobQueue.new()
         set :authentication, false
-        config_file = File.join(
-                File.expand_path(Dir.pwd),
-                "/config/api/config.yml"
-              )
+        config_file = File.join(Dir.pwd, "/config/api/config.yml")
         opts = YAML.load_file(config_file)
         opts["config_file"] = config_file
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -46,3 +46,7 @@ key.pem
 
 # API Database
 api.db
+
+# Config files
+bin/ssh_scan_api_example_config.yml
+bin/ssh_scan_worker_example_config.yml
```
claudijd: Good idea, this should prevent accidental config exposure.

lib/ssh_scan/constants.rb
```diff
@@ -67,8 +67,8 @@ module Constants
           "pwnbus",
           "kang",
         ],
-        :glitter => "https://gitter.im/mozilla-ssh_scan/Lobby",
-        :glitter_contacts => [
+        :gitter => "https://gitter.im/mozilla-ssh_scan/Lobby",
+        :gitter_contacts => [
```
claudijd: ha! here I was calling it glitter for a long time, thanks!

lib/ssh_scan/database/sqlite_db.rb
```diff
@@ -55,6 +57,17 @@ def find_scan_result(uuid)
...
+
+      def fetch_available_result(socket)
+        uuid = []
+        @db.execute(
+          "select uuid from api_schema where target like ( ? ) and port like ( ? )",
+          [socket[:target], socket[:port]]
+        ) do |row|
+          uuid << row[1]
+        end
+        return uuid
```
claudijd: I think in this interface you're returning the UUID as a string and not as JSON, where as in the other one, you return JSON.  This should be consistent between the two interfaces, if we expect them to be inter-changable.

rishabhs95: Yeah, I will change that, I still have to do more testing with mysql part

claudijd:
What about cache timeouts?

rishabhs95:
I have introduced caching timeouts and made the whole feature consistent with both sqlite and mongodb. The feature is working as expected on my system and is passing all the tests. Do have a look and let me know if there are any errors/inconsistencies. @claudijd @agaurav77 @royalharsh

lib/ssh_scan/api.rb
```diff
@@ -103,10 +104,23 @@ class API < Sinatra::Base
...
+              }.to_json
+            end
+          end
+        end
         options[:uuid] = SecureRandom.uuid
         settings.job_queue.add(options)
         {
-          uuid: options[:uuid]
+          uuid: options[:uuid],
+          scanned_on: 'fresh scan'
```
royalharsh: what's this about? I don't think we need to set any `scanned_on` property while adding it to the job queue.

rishabhs95: I'm not adding it here in the job queue. That's a return json block.
You will get the `scanned_on` timestamp along with `uuid` now.

royalharsh: I think we can just put the latest time instead of `fresh scan`?

rishabhs95: Can. Just that a tag like fresh scan or something looks more noticeable to me. Though I agree to that it may not be an accurate description to `scanned_on`,still tag instead of timestamp looked really straight forward in indicating that a fresh scan was initiated.

claudijd: I think there are two places we can handle this...

1.) We need a way to determine if a target is waiting in the JobQueue (there could be a method added to that object to check and if so, return the uuid to the end user)
2.) We need a way to determine the target has a recent scan in the Database. (there could be a method added to SSHScan::Database to check for pre-existing and recent scans from the target and return the UUID for that scan if it's within the cache timeout window.  This would also require respective interfaces added to MongoDB and SQLite objects that appear to operate the same from the call point of view.  Let me  know if you'd like some guidance here on that part.

royalharsh:
Coverage threshold for failure has been set at 80%.

rishabhs95:
I think the caching feature is ready. We can implement the feature to check if the task in the queue or not along with Issue #320. As otherwise, we will have to implement another in-memory store just to keep track of queued items, which would be much easier after DB integration.

lib/ssh_scan/api.rb
```diff
@@ -18,6 +18,13 @@ class API < Sinatra::Base
       configure do
         set :job_queue, JobQueue.new()
         set :authentication, false
+        config_file = File.join(
+                File.expand_path(Dir.pwd),
```
royalharsh: Only `Dir.pwd` will do, no need `File.expand_path`

rishabhs95:
@claudij What do you think about this?

lib/ssh_scan/api.rb
```diff
@@ -18,6 +18,13 @@ class API < Sinatra::Base
...
+                "/config/api/config.yml"
```
claudijd: should this be an absolute path (/) or a relative path (./)?

lib/ssh_scan/api.rb
```diff
@@ -103,10 +110,25 @@ class API < Sinatra::Base
         options[:sockets] <<
           "#{params[:target]}:#{params[:port] ? params[:port] : "22"}"
         options[:policy_file] = options[:policy]
+        options[:force] = params[:force] ? params[:force] : false
+        unless options[:force] == 'true'
+          available_result = settings.db.fetch_available_result(params)
```
claudijd: should you be passing params here or params[:target]?

lib/ssh_scan/api.rb
```diff
@@ -103,10 +110,25 @@ class API < Sinatra::Base
...
+          unless available_result.nil?
+            if (Time.now -
+                Time.parse(available_result[:scanned_on].to_s)
+               ) / (60 * 60 * 24) < 1
+              return {
+                uuid: available_result[:uuid],
+                scanned_on: available_result[:scanned_on]
```
claudijd: scanned_on is redundant in this context (the result itself has this information embedded)

lib/ssh_scan/api.rb
```diff
@@ -103,10 +110,25 @@ class API < Sinatra::Base
         options[:sockets] <<
           "#{params[:target]}:#{params[:port] ? params[:port] : "22"}"
         options[:policy_file] = options[:policy]
+        options[:force] = params[:force] ? params[:force] : false
+        unless options[:force] == 'true'
+          available_result = settings.db.fetch_available_result(params)
+          unless available_result.nil?
+            if (Time.now -
+                Time.parse(available_result[:scanned_on].to_s)
+               ) / (60 * 60 * 24) < 1
```
claudijd: This isn't all that clear what is going on?

Maybe make this into it's own helper method?

lib/ssh_scan/api.rb
```diff
@@ -103,10 +110,25 @@ class API < Sinatra::Base
...
+              }.to_json
+            end
+          end
+        end
         options[:uuid] = SecureRandom.uuid
         settings.job_queue.add(options)
         {
-          uuid: options[:uuid]
+          uuid: options[:uuid],
+          scanned_on: Time.now
```
claudijd: This isn't really accurate (it's merely just tasked), also the scan result contains start and stop times for actual network i/o style scan.

lib/ssh_scan/api.rb
```diff
@@ -158,12 +180,21 @@ class API < Sinatra::Base
...
+        socket[:port] = result['port']
 
         if worker_id.empty? || uuid.empty?
           return {"accepted" => "false"}.to_json
         end
 
-        settings.db.add_scan(worker_id, uuid, JSON.parse(request.body.first).first)
+        available_result = settings.db.fetch_available_result(socket)
+        unless available_result.nil?
+          settings.db.delete_scan(available_result[:uuid]) unless
```
claudijd: What's the reasoning behind deleting the scan from the database?

I would ideally like to keep this data around so I can query it for research/debugging/performance purposes.

lib/ssh_scan/database.rb
```diff
@@ -52,5 +52,10 @@ def delete_all
     def find_scan_result(uuid)
       @database.find_scan_result(uuid)
     end
+
+    # @return [Hash] result
+    def fetch_available_result(socket)
```
claudijd: I wonder if `fetch_cached_result` would be more clear here?

lib/ssh_scan/database/mongo.rb
```diff
@@ -26,9 +26,13 @@ def self.from_hash(opts)
...
-      def add_scan(worker_id, uuid, result)
-        @collection.insert_one("uuid" => uuid, "scan" => result,
-                          "worker_id" => worker_id, "scanned_on" => Time.now)
+       def add_scan(worker_id, uuid, result, socket)
+        @collection.insert_one("uuid" => uuid,
+                          "target" => socket[:target],
+                          "port" => socket[:port],
+                          "scan" => result,
+                          "worker_id" => worker_id,
+                          "scanned_on" => Time.now)
```
claudijd: I think scanned_on is redundant here, results should contain this value.

lib/ssh_scan/database/mongo.rb
```diff
@@ -47,6 +51,17 @@ def find_scan_result(uuid)
         
         return nil
       end
+
+      def fetch_available_result(socket)
+        results = @collection.find(:target => socket[:target], :port => socket[:port])
+        return nil if results.count.zero?
+        result = {}
+        results.each do |doc|
+          result[:scanned_on] = doc[:scanned_on]
```
claudijd: Same here, scanned_on redundancy

claudijd:
Looks pretty good, added some comments about where I think it could improve.

lib/ssh_scan/api.rb
rishabhs95: I needed, params[:target] and params[:socket] both for finding the available result, so I used params instead.

lib/ssh_scan/api.rb
rishabhs95: Oh, I thought to just keep a single scan for every target:port so that the first result will be the desired result.

But yes, I think that data can help us for debugging issues, I will change this.

rishabhs95:
I have modified the PR considering all the requested changes. Please take a look at the new changes @claudijd

rishabhs95:
I think the caching feature is ready and is working as expected. Please have a look if this looks okay @claudijd @pwnbus.

claudijd:
@rishabhs95 @royalharsh nice job guys! landed!

Issue #306, opened 2016-12-22
Provide a generic YAML loader functionality


claudijd:
- Should be able to load YAML files as config files
- It support the white-listing of config attributes and error if any config attribute is unexpected

Example of places we could use this, which implement their own loaders.

DB config loader
Policy loader
Worker config loader
API config loader

agaurav77:
So something like a `ConfigLoader` class which would work like
```
allowed_attrs_in_config_file = ["auth_methods", ...]
opts = ConfigLoader.load_config_and_warn(allowed_attrs_in_config_file)
```
Since the list of attributes would be too much sometimes, so lets store all attrs possible in a YAML file at `configs/api.yaml` (say), then
```
opts = ConfigLoader.load_config_and_warn(SSHScan::API)
```
where the class is sufficient to deduce the config file.

@claudijd @royalharsh @rishabhs95 ?

claudijd:
This is done, pending landing.

PR #307, opened 2016-12-29
add configuration loader + example

ashishgaurav13:
This is the idea I had for a generic configuration loader (#306). Let's discuss this and I'll add unit tests after this looks more mature.

Things I can improve
1. This doesn't allow nested whitelists. So if attribute A can have sub attribute B, there is no check for that. Only one level of whitelisting allowed.
2. I can condense the large number of attributes for the function to be taken as `opts` argument.
3. I still have no idea how to judge whether the syntax is YAML or JSON or XML for a given config string. Maybe I can do that better.

(Unit tests failing because one unit test relies on creating a tmpfile with no extension)

Thanks!

### DIFF
--- a/lib/ssh_scan/authenticator.rb
+++ b/lib/ssh_scan/authenticator.rb
@@ -7,7 +7,7 @@ module SSHScan
     end
 
     def self.from_config_file(config_file)
-      opts = YAML.load_file(config_file)
+      opts = SSHScan::ConfigLoader.load(config_file, self)
       SSHScan::Authenticator.new(opts)
     end
 
--- a/lib/ssh_scan/config_loader.rb
+++ b/lib/ssh_scan/config_loader.rb
@@ -1,0 +1,37 @@
+require 'logger'
+
+require 'ssh_scan/config_loader/yaml'
+
+module SSHScan
+  module ConfigLoader
+    def self.load(config, class_name = nil, logger = nil)
+      ext = File.extname(config) if config
+      case config
+      when File
+        if ext == ".yml"
+          loaded_config = SSHScan::ConfigLoader::Yaml.new(config, class_name, 'file', logger)
+        else
+          raise "no support for this kind of file yet: #{ext}"
+        end
+      when String
+        # Change for other formats when they are available?
+        if config.include?(".yml") && File.exist?(config)
+          if ext == ".yml"
+            loaded_config = SSHScan::ConfigLoader::Yaml.new(config, class_name, 'file', logger)
+          else
+            raise "no support for this kind of file yet: #{ext}"
+          end
+        else
+          if !!YAML.load(config)
+            loaded_config = SSHScan::ConfigLoader::Yaml.new(config, class_name, 'string', logger)
+          else
+            raise "unknown syntax"
+          end
+        end
+      else
+        raise "unsupported config object type #{config.class}"
+      end
+      return loaded_config.checkedOpts
+    end
+  end
+end
--- a/lib/ssh_scan/config_loader/yaml.rb
+++ b/lib/ssh_scan/config_loader/yaml.rb
@@ -1,0 +1,72 @@
+require 'yaml'
+
+module SSHScan
+  module ConfigLoader
+    class Yaml
+      def initialize(config, class_name = nil, mode = 'file', logger = nil)
+        @opts = case mode
+          when "file" then YAML.load_file(config)
+          when "string" then YAML.load(config)
+        end
+        @logger = logger || Logger.new(STDOUT)
+        @class_name = class_name ? class_name.name : ""
+        @good_opts = loadGoodOpts
+      end
+
+      def opts
+        @opts
+      end
+
+      def checkedOpts
+        unless @good_opts.nil?
+          allowed = @good_opts["allowed"]
+          error_count = 0
+          retrieved_opts = opts
+          retrieved_opts.each do |key, value|
+            unless allowed.include?(key)
+              @logger.error("attribute not allowed: #{key}")
+              error_count += 1
+            end
+          end
+          @logger.fatal("error count for #{@class_name} configuration: #{error_count}") if error_count > 0
+        end
+        @opts
+      end
+
+      def loadGoodOpts
+
+        # This function loads the appropriate yml file that lists the attributes allowed
+        # for a configuration file associated with a class. For example, suppose policy
+        # configuration files (SSHScan::Policy) can only have attributes a, b, c then
+        # we can have a whitelist at config/loader_configs
+        ###########
+        # allowed:
+        # - a
+        # - b
+        # - c
+        ###########
+        # Our requirement is one such whitelist per class, so a good way to name such files
+        # is just downcasing their classnames. Examples of class - whitelist filenames:
+        # SSHScan::Authenticator => config/loader_configs/authenticator.yml
+        # SSHScan::Policy        => config/loader_configs/policy.yml
+        # SSHScan::A::B::C       => config/loader_configs/a_b_c.yml
+        # and so on ... (SSHScan:: is redundant in these classnames, so let's not have it)
+
+        unless @class_name.nil?
+          configs_dir = File.join(File.dirname(__FILE__), "../../../config/loader_configs")
+          reduced_class_name = @class_name.downcase.split("::").join("_")
+          # 'sshscan_' in every class name is redundant
+          reduced_class_name.gsub!("sshscan_", "")
+          possible_file = File.expand_path(File.join(configs_dir, "#{reduced_class_name}.yml"))
+          begin
+            config = YAML.load_file(possible_file)
+          rescue Errno::ENOENT
+            @logger.error("File #{possible_file} does not exist.")
+          else
+            return config
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/ssh_scan/policy.rb
+++ b/lib/ssh_scan/policy.rb
@@ -1,4 +1,4 @@
-require 'yaml'
+require 'ssh_scan/config_loader'
 
 module SSHScan
   class Policy
@@ -17,12 +17,12 @@ module SSHScan
     end
 
     def self.from_file(file)
-      opts = YAML.load_file(file)
+      opts = SSHScan::ConfigLoader.load(file, self)
       self.new(opts)
     end
 
     def self.from_string(string)
-      opts = YAML.load(string)
+      opts = SSHScan::ConfigLoader.load(string, self)
       self.new(opts)
     end
   end
--- a/lib/ssh_scan/worker.rb
+++ b/lib/ssh_scan/worker.rb
@@ -27,7 +27,7 @@ module SSHScan
     end
 
     def self.from_config_file(file_string)
-      opts = YAML.load_file(file_string)
+      opts = SSHScan::ConfigLoader.load(file_string, self, @logger)
       SSHScan::Worker.new(opts)
     end
 
--- a/config/loader_configs/policy.yml
+++ b/config/loader_configs/policy.yml
@@ -1,0 +1,11 @@
+allowed:
+  - rules
+  - ruleset
+  - name
+  - ssh_version
+  - auth_methods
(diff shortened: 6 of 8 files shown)


lib/ssh_scan/authenticator.rb
```diff
@@ -7,7 +7,7 @@ def initialize(config)
     end
 
     def self.from_config_file(config_file)
-      opts = YAML.load_file(config_file)
+      opts = SSHScan::ConfigLoader.load(config_file, SSHScan::Authenticator)
```
claudijd: I think you can replace "SSHScan::Authenticator" with "self" in this context

lib/ssh_scan/config_loader.rb
```diff
@@ -0,0 +1,31 @@
...
+        when 'file'
+          ext = File.extname(config)
+          case ext
+            when ".yaml", ".yml"
+              loaded_config = SSHScan::ConfigLoader::Yaml.new(config, class_name, 'file', logger)
+              return loaded_config.checkedOpts
+            else
+              logger.warn("no support for this kind of file yet: #{ext}") if logger
+          end
+        when 'string'
```
claudijd: Consider making this an else statement, if it's anything but a yaml file, we want to hard exit in the current state.

lib/ssh_scan/config_loader.rb
```diff
@@ -0,0 +1,31 @@
...
+    def self.load(config, class_name = nil, mode = 'file', logger = nil)
+      case mode
+        when 'file'
+          ext = File.extname(config)
+          case ext
+            when ".yaml", ".yml"
+              loaded_config = SSHScan::ConfigLoader::Yaml.new(config, class_name, 'file', logger)
+              return loaded_config.checkedOpts
+            else
+              logger.warn("no support for this kind of file yet: #{ext}") if logger
```
claudijd: Would like to see this result in a hard failure if someone tried to load a config file that's not yaml.

lib/ssh_scan/config_loader.rb
```diff
@@ -0,0 +1,31 @@
...
+require 'ssh_scan/config_loader/yaml'
+
+module SSHScan
+  module ConfigLoader
+    def self.load(config, class_name = nil, mode = 'file', logger = nil)
+      case mode
+        when 'file'
+          ext = File.extname(config)
+          case ext
+            when ".yaml", ".yml"
```
claudijd: Let's consider standardizing on one ext format for yaml files.  I don't care which one, but let's pick one.

lib/ssh_scan/config_loader.rb
claudijd: Also, I think this looks for a string of "string".  If you want to match on a class type, you need to say String unquoted.

lib/ssh_scan/config_loader.rb
```diff
@@ -0,0 +1,31 @@
...
+              loaded_config = SSHScan::ConfigLoader::Yaml.new(config, class_name, 'file', logger)
+              return loaded_config.checkedOpts
+            else
+              logger.warn("no support for this kind of file yet: #{ext}") if logger
+          end
+        when 'string'
+          # If we require support for more configuration formats, this logic should
+          # be changed
+          case
+            when !!YAML.load(config)
```
claudijd: What's the purpose of "!!" in this context?

lib/ssh_scan/config_loader/yaml.rb
```diff
@@ -0,0 +1,53 @@
...
+          allowed = @good_opts["allowed"]
+          error_count = 0
+          retrieved_opts = opts
+          retrieved_opts.each do |key, value|
+            unless allowed.include?(key)
+              @logger.error("attribute not allowed: #{key}")
+              error_count += 1
+            end
+          end
+          @logger.error("error count for #{@class_name} configuration: #{error_count}") if error_count > 0
```
claudijd: Config loading errors should be hard failures, if we cannot load the config, there is no point in proceeding.

lib/ssh_scan/config_loader/yaml.rb
```diff
@@ -0,0 +1,53 @@
...
+      def loadGoodOpts
+        unless @class_name.nil?
+          configs_dir = File.join(File.dirname(__FILE__), "../../../configs")
+          reduced_class_name = @class_name.downcase.split("::").join("_")
+          # 'sshscan_' in every class name is redundant
+          reduced_class_name.gsub!("sshscan_", "")
+          possible_file = File.expand_path(File.join(configs_dir, "#{reduced_class_name}.yml"))
+          begin
+            config = YAML.load_file(possible_file)
+          rescue
```
claudijd: A bare rescue rescues any error that could possibly be thrown here.  If we're going to trap errors here, we should be explicit about what we're rescuing and if we don't know we should figure it out.

lib/ssh_scan/policy.rb
```diff
@@ -17,12 +17,12 @@ def initialize(opts = {})
     end
 
     def self.from_file(file)
-      opts = YAML.load_file(file)
+      opts = SSHScan::ConfigLoader.load(file, SSHScan::Policy)
```
claudijd: consider s/SSHScan::Policy/self/

lib/ssh_scan/policy.rb
```diff
@@ -17,12 +17,12 @@ def initialize(opts = {})
...
       self.new(opts)
     end
 
     def self.from_string(string)
-      opts = YAML.load(string)
+      opts = SSHScan::ConfigLoader.load(string, SSHScan::Policy, 'string')
```
claudijd: consider s/SSHScan::Policy/self/

lib/ssh_scan/config_loader/yaml.rb
```diff
@@ -0,0 +1,53 @@
+require 'yaml'
+
+module SSHScan
+  module ConfigLoader
+    class Yaml
+      def initialize(config, class_name = nil, mode = 'file', logger = nil)
```
claudijd: What is the reasoning for needing a logging attribute here?

Couldn't we just get the log information from the config? (I believe we already do this in some config examples, maybe we should just continue that pattern and loose this logger param)

configs/policy.yml
```diff
@@ -0,0 +1,11 @@
+allowed:
```
agaurav77: According to the naming scheme, any class SSHScan::A::B::C::D that requires configuration whitelisting can place a file `a_b_c_d.yml` in `configs`. This guarantees unique names for classes. (So SSHScan::Policy has `configs/policy.yml`)

I can think of a different scheme if you want.

lib/ssh_scan/config_loader.rb
agaurav77: Let's go with `.yml`

lib/ssh_scan/config_loader.rb
agaurav77: `!!YAML.load(string)` can ascertain whether `string` is valid YAML or not.

lib/ssh_scan/config_loader/yaml.rb
agaurav77: Agreed, this can be dropped. Will change.

lib/ssh_scan/config_loader.rb
```diff
@@ -0,0 +1,31 @@
+require 'logger'
+
+require 'ssh_scan/config_loader/yaml'
+
+module SSHScan
+  module ConfigLoader
+    def self.load(config, class_name = nil, mode = 'file', logger = nil)
```
claudijd: mode seems redundant in this context, this seems like something we can easily detect without requiring the user to specify.

Example:

    case config
    when File
       ...#it's a file
    when String
      if config.include?(".yml") && File.exist?(config)
        ... #it's a path
      else
        ... #it's a string
      end
    else
       raise "unsupported config object type #{config.class}"
    end

lib/ssh_scan/config_loader/yaml.rb
```diff
@@ -0,0 +1,53 @@
...
+        @opts = case mode
+          when "file" then YAML.load_file(config)
+          when "string" then YAML.load(config)
+        end
```
claudijd: Would like to lose this complexity by making ConfigLoader smarter

lib/ssh_scan/config_loader/yaml.rb
```diff
@@ -0,0 +1,53 @@
...
+          end
+          @logger.error("error count for #{@class_name} configuration: #{error_count}") if error_count > 0
+        end
+        @opts
+      end
+
+      def loadGoodOpts
+        unless @class_name.nil?
+          configs_dir = File.join(File.dirname(__FILE__), "../../../configs")
+          reduced_class_name = @class_name.downcase.split("::").join("_")
```
claudijd: I don't exactly understand why we're doing this.  Could you add more comments to make it more clear?

lib/ssh_scan/worker.rb
```diff
@@ -29,7 +29,7 @@ def setup_logger(logger)
     end
 
     def self.from_config_file(file_string)
-      opts = YAML.load_file(file_string)
+      opts = SSHScan::ConfigLoader.load(file_string, SSHScan::Worker, 'file', @logger)
```
claudijd: consider s/SSHScan::Policy/self/

claudijd:
Done with my initial round of review, let me know if you have any questions.

claudijd:
Per comment "I still have no idea how to judge whether the syntax is YAML or JSON or XML for a given config string. Maybe I can do that better."...

I don't think we need to worry about it, just yet.  I don't think we have an XML format.  YAML and JSON files should easily load into a common hash format, so you could just have a exception handling and try both, I think JSON and YAML are unique enough that they would respectively fail on JSON parser loading YAML and vise versa, so you do a first match strategy.

lib/ssh_scan/authenticator.rb
agaurav77: I think you mean `self.class`. `self` will be an instance, while I want the Class type.

lib/ssh_scan/config_loader/yaml.rb
agaurav77: This should be Errno::ENOENT

lib/ssh_scan/config_loader/yaml.rb
agaurav77: `ConfigLoader` won't require any of this, now that we're not having mode at all, since it can be guessed. 

I think it is okay to have this in internal files like `yaml.rb` that do this processing. I agree they are sort of "dumb", but the combined use makes `ConfigLoader` smart.

lib/ssh_scan/config_loader.rb
agaurav77: I was going to go with `logger.fatal`, but I think `raise` will be good enough.

lib/ssh_scan/config_loader/yaml.rb
agaurav77: Added

lib/ssh_scan/config_loader/yaml.rb
claudijd: I think we should drop logging altogether, if we run into an exception in config loading, it should be a hard failure.  I think this simplifies and sort eliminates this as an issue from a "what do we pass into this object to get logging?" problem

lib/ssh_scan/config_loader/yaml.rb
claudijd: I think we should have config loaded take a similar abstraction strategy as we did with the Database abstractions.  ConfigLoader should be able to be pointed as a file and figure it out, but we should have subclasses that do the heavy lifting (if any) for that particular config type format (JSON, YML, etc.)

lib/ssh_scan/config_loader/yaml.rb
pwnbus: I think this is a good idea, to fail fast and hard if we get an exception in the config.

claudijd:
please move this PR to https://github.com/mozilla/ssh_scan_api project

Issue #308, opened 2016-12-29
How to deal with multiple dependancy groups..


claudijd:
http://bundler.io/v1.5/groups.html

claudijd:
decided to break the project into multiple repos/gems

https://github.com/mozilla/ssh_scan_api

PR #310, opened 2017-01-08
Multiple dependency groups

rishabhs95:
Introduce bundler groups for lower core dependencies. Issue #308.

With this we will be able to do something like `bundle install --without (api/dev/mongo)` for just the core CLI scanner, and a regular `bundle install` for full product with API and everything.


claudijd:
@rishabhs95 I looked into this briefly, but this only solves the problem for source installations.  I'm beginning to think we really need to consider breaking ssh_scan_api out into it's own gem to avoid gem dependancy issues that we're seeing here.

claudijd:
I moved this code off into it's own repo, so we're no longer going to trip up on dependancies...

https://github.com/mozilla/ssh_scan_api

rishabhs95:
Yeah @claudijd, that looks like a better solution here :)

PR #311, opened 2017-01-11
Clean up db interfaces

claudijd:
Tightenes up some the DB interfaces and abstraction layers to make it flow more naturally and provide unit-tests for a lot of this functionality so it doesn't break in the future.


claudijd:
At this point, I have the updated DB interfaces working with the API using SQLite, need to work through the mongodb cases and add specs for that too, and we should be in pretty good shape.

claudijd:
@royalharsh @rishabhs95 @pwnbus let me know if you have any initial feedback good/bad about these mods.  I'll pivot toward making sure there is parity with mongodb and added unti-tests for it.

claudijd:
looks like @royalharsh was already on the war path with mongodb and api updates, I'm going to merge this code and then try another PR that incorporates his changes from a recent PR using the new format so it all jives well together.

royalharsh:
Cool! This is a lot cleaner than the previous implementation.

rishabhs95:
This is awesome. Should I rebase the API caching PR over this now or wait for the `scan_result` db consistency, as it is ready for review?

PR #313, opened 2017-01-11
Porting mongo/api specs to new DB interfacing

claudijd:


lib/ssh_scan/api.rb
```diff
@@ -113,17 +113,25 @@ class API < Sinatra::Base
...
-        settings.db.delete_scan(uuid)
+        if uuid.nil? || uuid.empty?
+          return {"deleted" => "false"}.to_json
+        else
+          scan = settings.db.find_scan_result(uuid)
+          if scan.empty?
+            return {"deleted" => "false"}.to_json
+          else
+            settings.db.delete_scan(uuid)
+            return {"deleted" => "true"}.to_json
```
claudijd: @royalharsh I added this, because the true case was never reporting a true feedback.

lib/ssh_scan/database/mongo.rb
```diff
@@ -2,36 +2,47 @@
...
-        @scans.delete_one(:uuid => uuid)
+        @collection.delete_one(:uuid => uuid)
       end
 
       def delete_all
-        @scans.delete_many({})
+        @collection.delete_many({})
       end
 
+      # LEFT OFF HERE: the results of this method should be the exact same format as with SQLite
```
claudijd: This is an area where this still needs work, the responses from find_scan_result should be the exact same format between mongodb and sqlite, they are not currently the case.  When this is fixed, the unit-tests can leverage shared specs so we know they will behave the same way.

claudijd: I will continue to work on this in an effort to unify the response behavior

PR #318, opened 2017-01-12
Testing coveralls

royalharsh:
Testing coveralls.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ matrix:
   - rvm: 2.3.0
     env:
       - LABEL=unit_tests
+    after_success:
+      - coveralls
   - rvm: 2.2.0
     env:
       - LABEL=unit_tests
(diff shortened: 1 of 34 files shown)


rishabhs95:
Is coveralls posting the correct coverage report here?
I think it's not summarizing the report, rather it is posting for every matrix component.

royalharsh:
Yeah, it is posting for every matrix component. But, the coverage is correct.

rishabhs95:
Yeah, I meant is it right to post 5 individual posts for every push?
The readme tag will work right but PR should only get 1 coverall comment.

royalharsh:
I know it's a bit noisy. One thing we can do is that we can run coveralls only on one version of ruby. Another thing we can do is that we can turn off commenting on pull requests.

rishabhs95:
Yeah, I think checking coverage on the latest ruby version should work as if the tests are passing coverage won't depend upon the ruby version.

claudijd:
@rishabhs95 I agree with the latest ruby approach, I think commenting, assuming it's only one comment per push is acceptable noise level.  Nice work here!

Issue #309, opened 2017-01-05
Expose an API route to gather queue/db/request statistics


claudijd:
- How many items in the queue? (from the queue)
- Average time for a worker to complete a job? (from the DB)
- How many errors?
- How many scan requests per some time period?
- How many scan requests total?

claudijd:
Route name ideas:

- /api/v1/health
- /api/v1/stats

claudijd:
this is done, /api/v1/stats

PR #322, opened 2017-01-26
Gather statistics

royalharsh:
Solves: https://github.com/mozilla/ssh_scan/issues/309

 - `/api/v1/stats`

### DIFF
--- a/lib/ssh_scan/api.rb
+++ b/lib/ssh_scan/api.rb
@@ -104,6 +104,13 @@ https://github.com/mozilla/ssh_scan/wiki/ssh_scan-Web-API\n"
           "#{params[:target]}:#{params[:port] ? params[:port] : "22"}"
         options[:policy_file] = options[:policy]
         options[:uuid] = SecureRandom.uuid
+        settings.count_req += 1
+        if (Time.now - settings.curr_time) > 60.0
+          settings.requests_per_min = 0
+          settings.curr_time = Time.now
+        else
+          settings.requests_per_min += 1
+        end
         settings.job_queue.add(options)
         {
           uuid: options[:uuid]
@@ -162,8 +169,25 @@ https://github.com/mozilla/ssh_scan/wiki/ssh_scan-Web-API\n"
         if worker_id.empty? || uuid.empty?
           return {"accepted" => "false"}.to_json
         end
+        results = JSON.parse(request.body.first).first
+        settings.current_avg = settings.current_avg +
+        (results['scan_duration_seconds'] - settings.current_avg) / settings.count_req
+        settings.db.add_scan(worker_id, uuid, results)
+      end
 
-        settings.db.add_scan(worker_id, uuid, JSON.parse(request.body.first).first)
+      get '/stats' do
+        time_spent = Time.now - settings.curr_time
+        if time_spent > 60.0
+          settings.curr_time = Time.now
+          settings.requests_per_min = 0
+        end
+
+        {
+          :items_queued => settings.job_queue.size,
+          :total_scan_requests => settings.count_req,
+          :avg_time_worker => settings.current_avg,
+          :requests_per_min => settings.requests_per_min
+        }.to_json
       end
 
       get '/__lbheartbeat__' do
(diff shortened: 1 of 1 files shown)


lib/ssh_scan/api.rb
```diff
@@ -184,6 +208,10 @@ def self.run!(options = {}, &block)
         set :logger, Logger.new(STDOUT)
         set :job_queue, JobQueue.new()
         set :db, SSHScan::Database.from_hash(options)
+        set :count_req, 0
+        set :current_avg, 0.0
+        set :requests_per_min, 0
+        set :curr_time, Time.now
```
rishabhs95: The name `:curr_time` seems confusing here, change to `:window_start_time` or something like that?

lib/ssh_scan/api.rb
```diff
@@ -162,8 +169,25 @@ class API < Sinatra::Base
         if worker_id.empty? || uuid.empty?
           return {"accepted" => "false"}.to_json
         end
+        results = JSON.parse(request.body.first).first
+        settings.current_avg = settings.current_avg +
+        (results['scan_duration_seconds'] - settings.current_avg) / settings.count_req
```
rishabhs95: You are assuming here that all the requests arrive at `:curr_time(:window_start_time)`, but what if the requests are uniformly distributed over the time period?

rishabhs95: Even if you want to do it this way, shouldn't be the `new_avg = (old_avg*n + time_spent_on_curr_task) / (n+1)`? ,i.e.,
`settings.current_avg = (settings.current_avg*(settings.count_req-1) + results['scan_duration_seconds'] - settings.current_avg) / settings.count_req`

lib/ssh_scan/api.rb
```diff
@@ -104,6 +104,13 @@ class API < Sinatra::Base
           "#{params[:target]}:#{params[:port] ? params[:port] : "22"}"
         options[:policy_file] = options[:policy]
         options[:uuid] = SecureRandom.uuid
+        settings.count_req += 1
+        if (Time.now - settings.curr_time) > 60.0
+          settings.requests_per_min = 0
+          settings.curr_time = Time.now
+        else
+          settings.requests_per_min += 1
+        end
```
claudijd: I think we should have a Ruby object class that tracks requests in a unified way.  I feel like that might make the integration a bit cleaner that splicing the raw code into the API route code as is.  It could simply act as a non-blocking gateway for new scan requests (Example `Stats.new_scan_request`) and it could internally manage all the attributes we care about and also be easily unit-tested outside of the API context. (Example `Stats.requests_per_min`, `Stats.avg_requests_per_min`, `Stats.total_requests`)

As for the stats that require information from the DB, we could add a method off of those to generate the stats for us or we could use the Stats interface to call the DB (Example `db.avg_worker_time`) and retrieve those details.  DB might be the most straight forward.

royalharsh: Yeah. It would be much cleaner.

lib/ssh_scan/api.rb
royalharsh: Actually, it's the same that you wrote.

lib/ssh_scan/api.rb
rishabhs95: Yeah, I agree.

lib/ssh_scan/api.rb
rishabhs95: Oh, I thought `results['scan_duration_seconds'] - settings.current_avg` is equal to `time_spent_on_curr_task`, whereas `results['scan_duration_seconds']` is.

Then, its correct. (y)

lib/ssh_scan/api.rb
claudijd: I'm going to send you a PR as a suggestion or start to make this a bit more clear.

claudijd: It would also be cool if eventually the DB could fully take over in terms of stats, so it will reduce our memory consumption on the API and make it leaner, but having the stats object still makes this cleaner in the interim, at least IMO.

royalharsh:
I think we can close this.

PR #323, opened 2017-01-26
Suggestion for stats interface

claudijd:
@royalharsh @rishabhs95 what do you think about something like this for cleaning up #322 ?

### DIFF
--- a/lib/ssh_scan/stats.rb
+++ b/lib/ssh_scan/stats.rb
@@ -1,0 +1,43 @@
+module SSHScan
+  class Stats
+    def initialize
+      @requests = []
+    end
+
+    def new_scan_request
+      @requests << Time.now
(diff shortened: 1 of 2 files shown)


lib/ssh_scan/stats.rb
```diff
@@ -0,0 +1,43 @@
...
+      @requests = []
+    end
+
+    def new_scan_request
+      @requests << Time.now
+    end
+
+    def size
+      @requests.size
+    end
```
claudijd: I guess you could also just add a method to say "give me all stats" as a hash and then just JSONify that and make the /api/stats route a one-liner :)

royalharsh:
This looks a lot cleaner and will be a good starting point for us. :smile:

Issue #320, opened 2017-01-12
Make JobQueue Database aware


claudijd:
Basic is that we would replace Ruby Queue object logic with actual database calls.

claudijd:
I think we have two options here:

1.) Make JobQueue DB aware by making job queue manage a queue table in each DB type
2.) DB smarter and allow it to add/remove from the queue table directly (preferred by @pwnbus and @claudijd)

Note: be careful about ordering, we want a first in, first out flow.  This may require DB specific logic underneath the hood.

claudijd:
Expanding on this.... We might be able to get away with a single table for queue and results.  The idea being that there could be state values described for each item.

**States could be:** pending, running, finished

Idea being that we could also measure time between states and understand performance/benchmark information.

PR #325, opened 2017-02-02
Make job_queue database-aware

rishabhs95:
I've tried to keep the functions similar to the Ruby's Queue class to make minimum changes to the code. Issue #320.

The PR is not ready yet but you can see and suggest if there should be any changes. Meanwhile I'll make some changes to make it better and functional with the current API. @claudijd @pwnbus @agaurav77 @royalharsh.

### DIFF
--- a/lib/ssh_scan/database/mongo_queue.rb
+++ b/lib/ssh_scan/database/mongo_queue.rb
@@ -1,0 +1,49 @@
+require 'mongo'
+
+Mongo::Logger.logger.level = ::Logger::FATAL
+
+module SSHScan
+  module QueueDB
+    class MongoDb
+      attr_reader :queue_collection
+
+      def initialize(queue_client)
+        @queue_client = queue_client
+        @queue_collection = @queue_client[:ssh_scan_queue]
+      end
+
+      # Helps us create a SSHScan::DB::MongoDB object with a hash
+      def self.from_hash(opts)
+        name = opts["queue_db_name"]
+        server = opts["server"]
+        port = opts["port"]
+        socket = server + ":" + port.to_s
+
+        queue_client = Mongo::Client.new([socket], :database => name)
+        return SSHScan::QueueDB::MongoDb.new(queue_client)
+      end
+
+      # @param [String] socket
+      def push(socket)
+        @queue_collection.insert_one("socket" => socket, "status" => "pending")
+      end
+
+      def size
+        @queue_collection.find("socket" => socket, "status" => "pending").count
+      end
+
+      def empty?
+        size.zero?
+      end
+
+      def pop
+        @queue_collection.find_one_and_update({ "socket" => socket, "status" => "pending" },
+                                        { "$set" => { "status" => "running" }})
+      end
+
+      def delete_all
+        @queue_collection.delete_many({})
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


lib/ssh_scan/database/mongo_queue.rb
```diff
@@ -0,0 +1,49 @@
...
+        port = opts["port"]
+        socket = server + ":" + port.to_s
+
+        queue_client = Mongo::Client.new([socket], :database => name)
+        return SSHScan::QueueDB::MongoDb.new(queue_client)
+      end
+
+      # @param [String] socket
+      def push(socket)
+        @queue_collection.insert_one("socket" => socket, "status" => "pending")
```
royalharsh: Maybe we can just add an extra field `status` in `ssh_scan.db` and be done away with a second db `ssh_scan_queue.db`?

rishabhs95: We are going to use the database to store scan results anyways, even if we switch job_queue later to something like redis, that is why I wanted to keep them de-coupled.

lib/ssh_scan/database/mongo_queue.rb
```diff
@@ -0,0 +1,55 @@
+require 'mongo'
+
+Mongo::Logger.logger.level = ::Logger::FATAL
+
+module SSHScan
+  module QueueDB
+    class MongoDb
+      attr_reader :queue_collection
```
claudijd: I think we'd be better off here if we leveraged the pre-existing DB classes and made JobQueue just use the high-level DB abstraction.

claudijd: SSHScan::JobQueue has methods that talk to SSHScan::DB

-OR-

SSHScan::DB has added methods for queuing and we drop SSHScan::JobQueue entirely

claudijd: My concern with the existing approach is that it created a lot of new code and maybe that's not needed for this additional function.  Really all we're doing is adding an additional table for queue info.

lib/ssh_scan/job_queue.rb
```diff
@@ -1,7 +1,28 @@
...
+    def self.from_hash(opts)
+      database_options = opts["database"]
+
+      # Figure out what database object to load
+      case database_options["type"]
+      when "sqlite"
+        queue_database = SSHScan::QueueDB::SQLite.from_hash(database_options)
+      when "mongodb"
+        queue_database = SSHScan::QueueDB::MongoDb.from_hash(database_options)
+      else
```
claudijd: by going the approach I mentioned above, we can skip having to do this again

claudijd:
Would like to propose we moved this PR over to https://github.com/mozilla/ssh_scan_api

Leaving PR open until we have an eqivalent one in the new repo.

rishabhs95:
@claudijd I will move the PR over to `ssh_scan_api` over this week and change the current implementation to a better implementation :)

claudijd:
Please move this PR to the ssh_scan_api project (https://github.com/mozilla/ssh_scan_api)

rishabhs95:
@claudijd I got caught up in some work... will move the PR over to the other repository before this weekend

rishabhs95:
Closing, as this PR has been migrated to `ssh_scan_api`

Issue #328, opened 2017-02-09
API authentication is broken


agaurav77:
Currently, scans are not processed if they are added via an authenticated request. For example, if I add a token `testpass` for user `test`, and do

```bash
$ curl --data "target=192.168.11.22&port=9999" http://localhost:8000/api/v1/scan -H "SSH_SCAN_AUTH_TOKEN: testpass"
{"uuid":"dc90e613-5c20-4de6-b893-44bc86d39b16"}
```

then I get a scan not found error when I try to retrieve the scan:

```bash
$ curl http://localhost:8000/api/v1/scan/results\?uuid\=dc90e613-5c20-4de6-b893-44bc86d39b16 -H "SSH_SCAN_AUTH_TOKEN: testpass"
{"scan":"not found"}
```

claudijd:
It's hard to say if this was fixed in my recent bug fix of the workers or you had it setup incorrectly.

I did spend some time tonight to setup an environment and verify the authentication stuff works against  current master branch (31f118f7d7bd241dd4e6ab5ecee8bc8b51440247) with some test tokens.  Here's the config I used...

		$ git diff
		diff --git a/config/api/config.yml b/config/api/config.yml
		index 1db633b..fc9aa2b 100644
		--- a/config/api/config.yml
		+++ b/config/api/config.yml
		@@ -7,7 +7,7 @@ port: 8000
		 # API Tokens (a crude mechanism for adding auth to your API)
		 #authentication: true
		-authentication: false
		+authentication: true
		 # If authentication is to true, the authentication tokens you generate below
		 # will be required to access certain API functions.
		@@ -15,17 +15,17 @@ authentication: false
		 # Generate a new token using Ruby's Secure random uuid
		 # $ ruby -r securerandom -e 'puts SecureRandom.uuid'
		 #
		-# users:
		-#   - username    : starlord
		-#     token       : INSERT_GROOT_TOKEN
		+users:
		+  - username    : starlord
		+    token       : c7b5a761-71a8-4814-8849-d2208247cf63
		 #   - username    : groot
		 #     token       : INSERT_GROOT_TOKEN
		 #   - username    : gamora
		 #     token       : INSERT_GAMORA_TOKEN
		 
		-# workers:
		-#   - worker_name : worker1
		-#     token       : INSERT_WORKER1_TOKEN
		+workers:
		+  - worker_name : worker1
		+    token       : 2463c915-7c35-48e6-bb79-7a03c6a15f94
[...]

Issue #321, opened 2017-01-13
Add authentication and authentication bypass for workers


claudijd:
When we enable authentication right now, all API name-spaced routes require authentication tokens.

This is cool and works well for the end user, but it creates a weird distribution problem for workers, especially a worker is on the same system that really shouldn't need a token.

This is what I'd like to propose:

1. No not require authentication from loopback sources (this includes users and workers and makes local operations/debugging more sane)
2. Implement authentication logic for the ./bin/ssh_scan_worker so that when a worker is setup remotely it knows how to properly authenticate.
3. Add a worker namespace under the authentication config that's separate from users.  Here we can generate tokens for workers and make sure they authenticate for every request they make.

agaurav77:
This will depend on #328.

PR #329, opened 2017-02-09
probable repair for auth

rishabhs95:
I think this may not be the best solution, but just posting this as it works. :)
Issue #328.

I think we can get a better solution with the suggestions from #321.


pwnbus:
Nice! Looks good!

PR #341, opened 2017-03-21
Update README.md

amgrice:
Corrected link to example polices to point to config/policies. No code change, just found that the link to the polices is now under `config` not just in the root. Wanted to updated the README.

- What did I do?
Clicked on link to example policies.
- What did I expect to happen?
To be taken to the example policies
- What happened instead?
I was taken to a Github 404


claudijd:
@amgrice thank you!

PR #349, opened 2017-03-24
Updates test for localhost resolution

petems:
* This test fails on a machine that does not 
resolve the ipv6 localhost address (eg. a machine
where /etc/hosts does not contain `::1 localhost`


claudijd:
@petems thanks for the patches! Much appreciated!

claudijd:
Fixes included in https://github.com/mozilla/ssh_scan/releases/tag/0.0.18

PR #356, opened 2017-03-26
Replace SQLite with YAML::Store

### LINKED ISSUE #355: Replace SQLite in ssh_scan with PStore/YAML::Store
We use SQLite3 in ssh_scan only for persistent host key fingerprint tracking, which is used to catch duplicate host-keys which are shared between targets.

This is cool for most OSs, but in cases like Fedora, this is slightly annoying in that a `gem install ssh_scan` can't be run without forcefully installing the gem or adding build tools to the target.

I found this good article which pretty much highlights the exact thought process I was headed down...

https://robm.me.uk/ruby/2014/01/25/pstore.html

This seems like a direction for eliminating SQLite all-together, which I like very much.

claudijd:
Fixes #355

### DIFF
--- a/bin/ssh_scan
+++ b/bin/ssh_scan
@@ -18,7 +18,7 @@ options = {
   "threads" => 5,
   "verbosity" => nil,
   "logger" => Logger.new(STDERR),
-  "fingerprint_database" => File.join(File.dirname(__FILE__),"../data/fingerprints.db")
+  "fingerprint_database" => File.join(File.dirname(__FILE__),"../data/fingerprints.yml")
 }
 
 # Reorder arguments before parsing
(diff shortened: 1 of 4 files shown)


bin/ssh_scan
```diff
@@ -246,6 +246,11 @@ ssh_scan' to get the latest"
   end
 end
 
+# Limit scope of fingerprints DB to (per scan)
```
claudijd: This is mainly for performance/useability reasons

1.) Performance - the file will just get larger and larger as you scan more stuff, and doing look ups will take longer as more fingerprints are added
2.) If you're a tester and you scan customer A (192.168.0.0/24) and you go to customer B (192.168.0.0/24) and they have the same appliance with the shared host key, this could get confusing because we're merging colliding network space.

Issue #396, opened 2017-06-26
Added new MACs and ciphers to SSH configuration, restarted server and ssh_scan fails


DJManas:
Hello,

I had used the ssh_scan to test my ssh config. It gave me some hints to add ciphers and MACs. I did it, tested config file, restarted ssh server and tried ssh_scan again and it fails.

Using ssh_scan version 0.0.21 using Docker on MacOS

Terminal output:
/app/lib/ssh_scan/result.rb:261:in `to_hash': undefined method `error=' for #<Hash:0x000000026a6fb0> (NoMethodError)
	from /app/lib/ssh_scan/scan_engine.rb:237:in `block in scan'
	from /app/lib/ssh_scan/scan_engine.rb:237:in `map'
	from /app/lib/ssh_scan/scan_engine.rb:237:in `scan'
	from /app/bin/ssh_scan:258:in `<main>'

I can give you the command, but I would give it to PM not public.

Thanks,
Regards,
Petr Sourek

claudijd:
@DJManas thanks for the report and sorry for the delay in response here.  Would you mind sending me a copy of your config file to claudijd@yahoo.com so I can reproduce the bug?

claudijd:
@DJManas I'm digging into some possible error cases in result management and will uplift some added unit-testing and better behavior on this soon.  Also, you may want to try the latest docker image with 0.0.23 and see if that helps, I may have already fixed this in a previous case.  In either case, I hope that this won't be an issue soon, but would love your feedback on this so we can get it sorted.

claudijd:
I think I've found another issue, which is causing docker builds not to update, so this could by why you were stuck with dealing with this bug, which I believe was fixed in code.  I'll send another PR to address that and you should be able to re-test after that.

claudijd:
Reverted some config that was unintentionally blocking docker pushes in https://github.com/mozilla/ssh_scan/pull/399

claudijd:
@DJManas could you please do the following on your end and let me know if you're all fixed up?

docker pull mozilla/ssh_scan
docker run -it mozilla/ssh_scan /app/bin/ssh_scan -t INSERT_YOUR_SERVER_NAME

DJManas:
Hello, I am sorry that I was not responding last week, I was on vacation, with no internet connection (great place, but this is drawback, when I had reported bug :-()

I had upgraded to latest version and it works. Thank you for your efford and again, sorry for not communicating.

Regards,
Petr Sourek

[...]

PR #398, opened 2017-07-03
Add unit-tests for potential results bug

claudijd:
Validates #396

### DIFF
--- a/lib/ssh_scan/result.rb
+++ b/lib/ssh_scan/result.rb
@@ -142,12 +142,8 @@ module SSHScan
     end
 
     def scan_duration
-      if start_time.nil?
-        raise "Cannot calculate scan duration without start_time set"
-      end
-
-      if end_time.nil?
-        raise "Cannot calculate scan duration without end_time set"
+      if start_time.nil? || end_time.nil?
+        return nil
       end
 
       end_time - start_time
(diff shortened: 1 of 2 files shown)


lib/ssh_scan/result.rb
```diff
@@ -142,12 +142,8 @@ def set_end_time
     end
 
     def scan_duration
-      if start_time.nil?
-        raise "Cannot calculate scan duration without start_time set"
-      end
```
claudijd: I thought having exception handling here was going to be helpful, but all it did was make it harder to spec out the results object, so I'm removing it.

spec/ssh_scan/result_spec.rb
```diff
@@ -212,4 +215,22 @@
...
+        :compliant => true,
+        :recommendations => ["do this", "do that"],
+        :references => ["https://reference.example.com"],
+      }
+      result = SSHScan::Result.new()
+      result.set_compliance = compliance
+      result.error = "This is an error"
+
+      expect(result.to_hash).to be_kind_of(Hash)
+      expect(result.to_hash["error"]).to eql("This is an error")
```
claudijd: I believe this was the case that would have triggered in issue #396, but was fixed between 0.0.21 and 0.0.23

PR #407, opened 2017-07-20
Make attribute comparisons blind to specific sshlib implementation tags

### LINKED ISSUE #405: Cipher compliance should be implementation independant (cipher == cipher@implementation => true)
Just scanned a machine and it recommends "Remove these key exchange algorithms: curve25519-sha256". Since this is probably the best one, it should not be removed. I guess this happens because ssh_scan is only aware of `curve25519-sha256@libssh.org` (including the @... part).

claudijd:
Fixes #405 (or will fix when I'm done with it)


claudijd:
I realized this affects more than just cipher attributes, so I'm making a generic capability for attribute comparison and will implement that in the policy manager.

Issue #406, opened 2017-07-19
Add support to scan .onion addresses


claudijd:

claudijd:
@yawnbox any chance you have an example .onion ssh service I could use for testing? If so, that might speed up getting this feature landed because it's less work for me to set one up.

claudijd:
# How to connect with an onion service via ncat
/usr/bin/ncat -v --proxy localhost:9050 --proxy-type socks5 3g2upl4pq6kufc4m.onion 80

# We can also setup proxy chains to force communications to go through TOR
proxychains /usr/bin/ncat -v 3g2upl4pq6kufc4m.onion 80

claudijd:
I'm thinking that it would be best if we just provided a SOCKS proxy option for ssh_scan binary as a switch will be the most direct way of handling/testing this.

We can also patch in TOR+ProxyChains into any Docker deployment we have as dependencies and let the end user of the container choose whether they want to start up those services via the endpoint.

By default any attempts to scan a .onion will require a SOCKS proxy and thus will throw an exception if one is not provided.

That's basically my approach here, welcome feedback, but I think this will work.

claudijd:
Looks like Net::SSH has SOCKS5 support here (that's 50% of what we need):

http://www.rubydoc.info/github/net-ssh/net-ssh/Net/SSH/Proxy/SOCKS5

claudijd:
Looks like we can use ruby-proxifier for the ruby socket calls in the SSHScan::Client (that's the other 50%):

https://github.com/samuelkadolph/ruby-proxifier

claudijd:
Here are the things that are needed to get core ssh_scan to support this...

1.) Expose a --proxy option on ./bin/ssh
2.) Make SSHScan::Client SOCK5 capable to consume the option that would be passed (add ruby-proxifier shim)
3.) Make Net-SSH calls use the SOCK5 proxy client when supplied (no additional deps here)

claudijd:
Once the above are done, the docker and deployment specifics are mostly configuration at that point.

yawnbox:
@claudijd go ahead and try 2vytis5xf5djnaoo.onion

yawnbox:
Just a note, for ssh_config, you can use:

proxyCommand ncat -v --proxy localhost:9050 --proxy-type socks5 %h %p

doesn't have to list a specific host and port.

claudijd:
[...]

PR #409, opened 2017-07-21
Add initial socks/.onion support to ssh_scan

claudijd:
This allows ssh_scan to scan through a socks proxy.  It's slower (like 3x slower using TOR), but it works.


claudijd:
This is inspired by #406

claudijd:
decided to forego this level of complexity at this time, though now we know it's not that hard to do when we're ready for it.

PR #424, opened 2018-01-19
Bind to a specific version of netaddr

claudijd:
I ran into some problems with a deployment about netaddr not installing, I happened to notice I was rolling with netaddr 1.5.1 and 2.x was released today, so I think that's somehow related.  I may go and report that upstream once I get my thing fixed.


claudijd:
According to https://github.com/dspinhirne/netaddr-rb this is totally incompatible, so this is probably a good move for now, until we can officially port to the 2.x way of doing things.

PR #440, opened 2018-10-26
Use alpine build for Docker

netsgnut:
This PR proposes switching to an Alpine-based image base `ruby:alpine` instead of using the default Debian-based `ruby` build. The switch can allow the general users to enjoy a smaller Docker image.

On my testing box, the image shrank from 913MB to 93.1MB.


claudijd:
@netsgnut cool, thanks!

PR #479, opened 2019-01-17
Bump for pre-release testing to fix #477

### LINKED ISSUE #477: no implicit conversion of nil into String (TypeError)
Since the upgrade to 0.0.37, I get the following error when running `ssh_scan -t mytarget`:

```
#<Thread:0x00007f84e41cd930@~/.gem/gems/ssh_scan-0.0.37/lib/ssh_scan/subprocess.rb:10 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
	2: from ~/.gem/gems/ssh_scan-0.0.37/lib/ssh_scan/subprocess.rb:16:in `block (3 levels) in initialize'
	1: from ~/.gem/gems/ssh_scan-0.0.37/lib/ssh_scan/scan_engine.rb:129:in `block in scan_target'
~/.gem/gems/ssh_scan-0.0.37/lib/ssh_scan/scan_engine.rb:129:in `+': no implicit conversion of nil into String (TypeError)
```

However, the scan is performed regularly and I can see the result right after the error.

claudijd:


claudijd:
The only reason this is failing is because the pre-release format doesn't conform to our semantic expectations, this is ok because once the pre-release is validated, I'll simply switch it to a real release.

PR #502, opened 2019-05-03
Up unit-test bounds so it's not failing on PR submits

claudijd:


claudijd:
I'm not going to roll a gem release for this because it just fixes a unit-test expectation.  Though it will be included in master and docker builds.

PR #511, opened 2020-01-02
Fix 510: tweak Dockerfile to work around upstream environment issue with Ruby 2.7.0

### LINKED ISSUE #510: Docker container fails with example in README
**Steps**

1. Pull the latest container (image ID `ab59549aae7c`) from DockerHub:
        
        $ docker pull mozilla/ssh_scan
        Using default tag: latest
        latest: Pulling from mozilla/ssh_scan
        Digest: sha256:5a039264d6786d1c0841252d2990e82bced5339d1e8f23b3cb04ec1fb7655b77
        Status: Image is up to date for mozilla/ssh_scan:latest
        docker.io/mozilla/ssh_scan:latest

1. Run the example command from the `README` file:

        $ docker run -it mozilla/ssh_scan /app/bin/ssh_scan -t sshscan.rubidus.com
        Traceback (most recent call last):
               2: from /app/bin/ssh_scan:7:in `<main>'
        	1: from /usr/local/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
        /usr/local/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require': cannot load such file -- netaddr (LoadError)

**Expected results**
SSH scan runs and returns JSON output.

**Actual results**
See traceback above.

**Technical**
I was able to get `ssh_scan` to work by entering the container via a shell, then manually running `gem install` for each of its dependencies. I suspect something about the gem installation process is broken for the container build.

rrazor:
## Overview

This PR resolves #510 by working around the non-apk-provided ruby 2.7.0 from `alpine:ruby` using environment variables.

I did the following:

- Redefined `GEM_HOME` from upstream to be accurate
- Removed `gem install bundler` as it is provided upstream in `/usr/local/bin`
- Added `build-base` to `apk del` to be symmetric with the temporary `apk add`

## Testing

There are no automated tests but I have manually verified that the following results in JSON output:

    $ docker build -t ssh_scan .
    $ docker run -it ssh_scan /app/bin/ssh_scan -t sshscan.rubidus.com
    [...]

## Additional steps

A better way to fix this may be to submit an upstream PR to the maintainers of `alpine:ruby` so that they can set the correct `GEM_*` environment variables according to the source-built version of ruby they choose to use.

For now, however, I believe this will correct the issue as reported.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,10 +9,12 @@ WORKDIR /app
 # required for ssh-keyscan
 RUN apk --update add openssh-client
 
-RUN apk --update add --virtual build-dependencies ruby-dev build-base && \
-    gem install bundler && \
+ENV GEM_HOME /usr/local/bundle/ruby/2.7.0/
+ENV GEM_PATH /usr/local/bundle/ruby/2.7.0/
+
+RUN apk --update add --virtual build-dependencies build-base && \
     bundle install && \
-    apk del build-dependencies && \
+    apk del build-dependencies build-base && \
     rm -rf /var/cache/apk/*
 
 CMD /app/bin/ssh_scan
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ WORKDIR /app
 # required for ssh-keyscan
 RUN apk --update add openssh-client
 
-ENV GEM_HOME /usr/local/bundle/ruby/2.7.0/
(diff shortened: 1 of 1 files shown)


claudijd:
Thanks for the fix!

PR #516, opened 2020-02-07
use latest stable net-ssh 5.2.0

mariuz:
use latest net-ssh https://github.com/net-ssh/net-ssh/blob/master/CHANGES.txt#L10


claudijd:
thanks @mariuz

PR #519, opened 2020-05-27
Add ed25519 and bcrypt_pbkdf per net-ssh#478

claudijd:


claudijd:
This is blocked by https://github.com/mozilla/ssh_scan/pull/513, unless we set an upper limit on net-addr and press on without it.

claudijd:
This is landed in ssh_scan v0.0.43

PR #525, opened 2020-12-02
Dockerfile improvements

J-M0:
I made a small tweak to the Dockerfile to make running ssh_scan from a container easier. Now instead of

```
docker run -it mozilla/ssh_scan /app/bin/ssh_scan -t sshscan.rubidus.com
```

it's just

```
docker run -it mozilla/ssh_scan -t sshscan.rubidus.com
```


claudijd:
Thanks @J-M0 !

